PR #4, opened 2020-10-22
PIP install

aslepakurov:
[SAT-62] [SAT-39]

- Pip package for starlark interpreter
- input/output parameters

[SAT-62]: https://verygoodsecurity.atlassian.net/browse/SAT-62
[SAT-39]: https://verygoodsecurity.atlassian.net/browse/SAT-39

### DIFF
--- a/pylarky/larky.py
+++ b/pylarky/larky.py
@@ -1,8 +1,0 @@
-import subprocess
-
-
-def evaluate(script, input_file, output_file):
-    print(subprocess.call(['larky-runner',
-                           '-input', input_file,
-                           '-output', output_file,
-                           script]))
--- a/pylarky/larky.py
+++ b/pylarky/larky.py
@@ -1,0 +1,8 @@
+import subprocess
+
+
+def evaluate(script, input_file, output_file):
+    print(subprocess.call(['larky-runner',
+                           '-input', input_file,
+                           '-output', output_file,
+                           script]))
--- a/pylarky/starlarky.py
+++ b/pylarky/starlarky.py
@@ -8,9 +8,11 @@ OUTPUT_PARAM = '-output'
 
 def evaluate(script, input_data: str) -> str:
     output_file = tempfile.NamedTemporaryFile(mode='w+')
-    input_path = __prepare_input_file(input_data)
+    input_file = tempfile.NamedTemporaryFile(mode='w+')
+    input_file.write(input_data)
+    input_file.flush()
 
-    __evaluate(script, input_path, output_file.name)
+    __evaluate(script, input_file.name, output_file.name)
 
     return open(output_file.name, mode='r').read()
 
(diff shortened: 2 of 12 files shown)


pylarky/larky.py
```diff
@@ -0,0 +1,8 @@
+import subprocess
+
+
+def evaluate(script, input_file, output_file):
```
rozenbom, line 4: IMO passing input data via file is a bit leaky abstraction and we should hide files usage behind the interface of this library passing an input data as a string/bytes. So preparing the input file and reading the output one will be responsibility of this library instread of Satellite:
```
def evaluate(script: str, input_data: str) -> str:
   pass
```

pylarky/larky.py
```diff
@@ -0,0 +1,8 @@
...
+    print(subprocess.call(['larky-runner',
```
rozenbom, line 5: 1. Would be nice to have some exception(s) exposed for the cases when something goes wrong during the evaluation process (larky-runner exited with non 0 status code).
2. We probably want to read stderr so the exception mentioned above would have some info that can be used by clients of this lib.

aslepakurov:
Python lib is locked on version `0.0.2` and will be rolling in further PR

pylarky/starlarky.py
```diff
@@ -0,0 +1,30 @@
+import tempfile
+import subprocess
+
+RUNNER_EXECUATBLE = 'larky-runner'
+INPUT_PARAM = '-input'
+OUTPUT_PARAM = '-output'
+
+
+def evaluate(script, input_data: str) -> str:
```
rozenbom, line 9: If we don't close (destroy) these files here they will live until a client app (Satellite) is existed which is not good.

pylarky/starlarky.py
```diff
@@ -0,0 +1,30 @@
...
+    return open(output_file.name, mode='r').read()
+
+
+def __evaluate(script, input_path, output_path):
+    code = subprocess.call([RUNNER_EXECUATBLE,
+                            INPUT_PARAM, input_path,
+                            OUTPUT_PARAM, output_path,
+                            script])
+    if code != 0:
+        raise FailedEvaluation(f'Evaluation returned non-zero return code. Code: {code}')
```
rozenbom, line 26: IMO exit code gives not much info about an error. Please consider reading stderr (or stdout if needed) of larky-runner and inject it into the exception message.

pylarky/starlarky.py
```diff
@@ -0,0 +1,35 @@
...
+    input_file.write(input_data)
+    input_file.flush()
+
+    __evaluate(script, input_file.name, output_file.name)
+    output = open(output_file.name, mode='r').read()
+
+    input_file.close()
+    output_file.close()
+
+    return output
```
rozenbom, line 21:
```suggestion
    with tempfile.NamedTemporaryFile(mode='w+') as output_file, tempfile.NamedTemporaryFile(mode='w+') as input_file:
        input_file.write(input_data)
        input_file.flush()
        __evaluate(script, input_file.name, output_file.name)
        return output_file.read()
```

pylarky/starlarky.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def __evaluate(script, input_path, output_path):
+    try:
+        check_output([RUNNER_EXECUATBLE,
+                      INPUT_PARAM, input_path,
+                      OUTPUT_PARAM, output_path,
+                      script], stderr=STDOUT)
+    except CalledProcessError as e:
+        raise FailedEvaluation(f'Starlark evaluation failed. \nOutput: {e.output}')
```
rozenbom, line 24:
```suggestion
        raise FailedEvaluation(f'Starlark evaluation failed. \nOutput: {e.output}') from e
```

rozenbom (approved):

PR #5, opened 2020-10-22
[ready4merge] simplification of Larky, rebased to work on top of update-starlark branch

mahmoudimus:

PR #7, opened 2020-10-23
Update libstarlark from upstream bazel (github.com/bazelbuild/bazel)

mahmoudimus:


mahmoudimus:
@aslepakurov this is ready for merge I believe. The failing tests will be fixed in #5

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "/Users/mahmoud/src/sbxscripting/bazelbuild/starlarky/.tmp/bazel"]
+	path = /Users/mahmoud/src/sbxscripting/bazelbuild/starlarky/.tmp/bazel
```
aslepakurov, line 2: `mahmoud` ?

aslepakurov: @mahmoudimus

mahmoudimus: Oh hmm, yes, I need to keep this relative path, instead of absolute. Does that work?

aslepakurov: yeap, that should work

aslepakurov (approved):

.gitmodules
aslepakurov: I will be merging this, can you please fix it in #5?

PR #12, opened 2020-11-08
Introduce Larkyloader

mahmoudimus:
- This allows us to load modules explicitly instead of having them implicitly via namespaces.
- Added `unittest.expectedFailure` to test failing assertions. Copied from [Python’s unittest](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure)
- Introduced the beginning of a JSR223 compliant script engine


mahmoudimus:
@aslepakurov merge #16 first here and then merge into master to preserve commit history, otherwise you will have to rebase.

aslepakurov (approved):

PR #13, opened 2020-11-10
WIP: grcp_service

criloz:
the request  need to define a `main` function and currently only support a map<string, string> for the context and payload only str, the other starlark types can be add with more detailed schema

```proto
message StarlarkRequest {
  string script = 1;
  string payload = 2;
  string main_fn = 3;
  map<string, string> ctx = 7;
}


message StarlarkResponse {
  string payload = 1;
  map<string, string> ctx = 2;
}


service ComputeService {
  rpc Compute (StarlarkRequest) returns (StarlarkResponse);
}
```

tranform the script from
```
def process(payload, ctx):
    ctx['foo'] = 'baz'
    return payload[::1]
```

to
```
def process(payload, ctx):
    ctx['foo'] = 'baz'
    return payload[::1]
    
def main_01EPS9BNB91MD3EPCAT3D2RAD2(inner):
    return {"payload": process(inner["payload"], inner["ctx"]), "ctx": inner["ctx"]}
main_01EPS9BNB91MD3EPCAT3D2RAD2(inner_ctx_01EPS9BNB9Y9KZH646CGZGH0RS)
```

where the `_ids` are generated randomly to avoid name clashing in the script scope

### DIFF
--- a/larky_server/protos/larky.proto
+++ b/larky_server/protos/larky.proto
@@ -16,6 +16,6 @@ message StarlarkResponse {
 }
 
 
-service ComputeService {
+service LarkyRuntimeService {
   rpc Compute (StarlarkRequest) returns (StarlarkResponse);
 }
--- a/larky_server/protos/larky.proto
+++ b/larky_server/protos/larky.proto
@@ -1,0 +1,21 @@
+syntax = "proto3";
+
+option java_package = "com.verygood.security.messages.operations.starlark";
+
+message StarlarkRequest {
+  string script = 1;
+  string payload = 2;
+  string main_fn = 3;
+  map<string, string> ctx = 7;
+}
+
+
+message StarlarkResponse {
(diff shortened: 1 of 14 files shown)


larky_server/protos/larky.proto
```diff
@@ -0,0 +1,21 @@
...
+}
+
+
+message StarlarkResponse {
+  string payload = 1;
+  map<string, string> ctx = 2;
+}
+
+
+service ComputeService {
```
Agitolyev, line 19: Hi, it is great to see the starlark server coming together :)
Please, consider renaming this service something more specific, like LarkyRuntimeService. We have a lot of stuff prefixing with Compute already and I fear that it will mislead people and overload the term.

criloz: @Agitolyev done!

PR #25, opened 2020-11-25
make  flexible the larky executable path

criloz:
- allow optionally to define the larky executable location and arguments, using environmental variables

- change `data_files`  keyword on setup.py for `package_data`, `data_files` is deprecated https://setuptools.readthedocs.io/en/latest/references/keywords.html

- add a bin folder under pylark where the binary should be located before run `setup.py`

- instead of call `larky-runner` directly it now locate the resource using the `pkg_resources` module to make it easy to distribute. based on this discussion: https://stackoverflow.com/questions/39104/finding-a-file-in-a-python-module-distribution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ jobs:
     steps:
       - checkout
       - <<: *attach-workspace
-      - run: cp ./larky/target/larky-runner .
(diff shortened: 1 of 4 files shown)


mahmoudimus:
@rozenbom @aslepakurov are we merging this?

aslepakurov:
outdated

PR #31, opened 2020-12-08
Http request for pylib

aslepakurov:
- Verbose merged script
- Http request parse
- Tests


rozenbom (commented):
Typo https://github.com/verygoodsecurity/starlarky/blob/1303b0195f5385f1eddf958e79b928a0170f4bcf/pylarky/eval/evaluator.py#L4

pylarky/eval/http_evaluator.py
```diff
@@ -1,12 +1,21 @@
...
 from pylarky.eval.evaluator import Evaluator
-from pylarky.request import HttpRequest
+from pylarky.model.request import HttpRequest
 
 
 class HttpEvaluator(Evaluator):
 
-    def __init__(self, script, http_request: HttpRequest):
-        super(HttpEvaluator, self).__init__(script, http_request.to_starlark())
+    def __init__(self, script):
```
rozenbom, line 10: BTW why do you need __init__ here at all?

setup.py
```diff
@@ -1,10 +1,18 @@
+import os
 from setuptools import find_packages, setup
 
+
+version = 'nover'
+try:
+    version = os.environ['CIRCLE_TAG']
```
rozenbom, line 7: Not sure that hardcoding CI-specific staff is the right way

mahmoudimus (approved):
This looks good to me

pylarky/eval/http_evaluator.py
```diff
@@ -1,12 +1,21 @@
...
+        super(HttpEvaluator, self).__init__(script)
+
+    def evaluate(self, http_request: HttpRequest) -> HttpRequest:
```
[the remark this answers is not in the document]
aslepakurov: I will change naming, as the structure of request should be the same - payload and headers

pylarky/eval/http_evaluator.py
aslepakurov: For possible future usage of using same script for multiple payloads

PR #42, opened 2021-01-22
[WIP] LAR-46 LAR-47 LAR-48 Sphinx starlark poc: autodoc

rozenbom:
## Fixes https://verygoodsecurity.atlassian.net/browse/LAR-48

This is POC for making Sphinx autodoc standard extension work with star-modules. Basically it tricks autodoc by converting a star-module into Python module (stripping out "load"-directive) on the fly and manipulating "sys.path".

For star-module like:
```python
"""
This is foobaz module
"""

# load should not break autodoc
load("@stdlib/larky", "larky")

def make_name(f_name, l_name):
    """
    Makes a name

    :param f_name: First name
    :type f_name: str
    :param l_name: Last name
    :type l_name: str
    """
    return f_name + ' ' + l_name
```

and rst-doc like:
```
Starlark Docs POC
=================

.. automodule:: foobaz
   :members:
```

you get something like this: 
Screenshot 2021-01-22 at 14 55 15

The initial commit in this PR has the original autodoc-code and the following commits do the alterations - so it's possible to tac what changes were made.


mahmoudimus:
We can probably avoid the larkyhack if we figure out a way to run a `larky2python` script on larky. Something to keep in mind, but that could be easier to deal w/ and I would expect our customers to be able to test larky scripts in their own python interpreter.

rozenbom:
> We can probably avoid the larkyhack if we figure out a way to run a `larky2python` script on larky. Something to keep in mind, but that could be easier to deal w/ and I would expect our customers to be able to test larky scripts in their own python interpreter.

I just assumed that we don't want to do this conversion step and let the sphinx extension do the magic.

blaisep-vgs:
OK, so there are two ways Sphinx can automatically generate docs:
autodoc: uses docutils to read nodes on the compiler's AST. Python only.
auto-api: a recent invention from the readthedocs crew, it parses the code
and can read several languages.

https://sphinx-autoapi.readthedocs.io/en/latest/how_to.html#how-to-transition-to-manual-documentation

If push comes to shove, we can use auto-api and write a parser for
Starlarky.

rozenbom:
> OK, so there are two ways Sphinx can automatically generate docs: autodoc: uses docutils to read nodes on the compiler's AST. Python only. auto-api: a recent invention from the readthedocs crew, it parses the code and can read several languages. https://sphinx-autoapi.readthedocs.io/en/latest/how_to.html#how-to-transition-to-manual-documentation If push comes to shove, we can use auto-api and write a parser for Starlarky.

The Autoapi-approach could fly especially considering that it uses astroid for Python and writing Starlark "parser" should be [trivial](/phgn0/pylint-starlark-plugin/blob/master/pylint_starlark_plugin/starlarkPlugin.py). However, autoapi approach is from the same realm as this PR - teaching some Sphinx plugin to work with Starlark. If we want to convert Starlark to Python anyway then all this hassle is unnecessary I guess and we can use vanilla Shpinx plugins with generated Python code.

blaisep-vgs:
Hi Petro,

SGTM, I have no opinion and I'm happy to facilitate whichever approach you
take.
For example, if we stick with Python and use the docutils description of
the nodes on the AST,
then we could use the vast collection of Sphinx plugins.
AND, if we needed to, we could write a custom sphinx "domain" to handle the
representations of Larky AST nodes and rST directives.
The Sphinx `python` domain is an example of such an enhancement
<https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-python-domain>
.

aslepakurov (approved):

rozenbom:
@mahmoudimus do you expect any further work from me here?

mahmoudimus:
@rozenbom thanks for the update here. This looks good, what's your plan on putting it in our normal documentation?

blaisep-vgs:
@rozenbom , if @mahmoudimus wants to include it in the main documentation,
there is a project that might help you,
https://pypi.org/project/pydoc-markdown/ will parse the code and produce
markdown (which seems to be prevalent in our current doc toolchain.

rozenbom:
> @rozenbom thanks for the update here. This looks good, what's your plan on putting it in our normal documentation?

Can you please clarify what do you mean by "normal documentation"? Moving **docs_poc** to **docs** :) ?
Also, I see there is another related PR from @blaisep-vgs  https://github.com/verygoodsecurity/starlarky/pull/43

la-luo:
new layout
Screen Shot 2021-12-07 at 4 19 25 PM

PR #43, opened 2021-02-03
Create (minimal) Sphinx project using the auto-api approach.

blaisep-vgs:
Preview at: https://starlarky-devel.rtfd.io

@mahmoudimus , this is an example of using Sphinx with the "parsing" approach [which can be easily adapted to languages beyond Python](https://sphinx-autoapi.readthedocs.io/en/latest/tutorials.html#python).

In this case I also juiced up the HTML builder by using a [sweet Sphinx theme called furo](https://pradyunsg.me/furo/kitchen-sink/)
One could also output to confluence, readthedocs, etc, but I opted to keep this very simple for clarity.

@rozenbom , let me know if you find this useful.


blaisep-vgs:
@aslepakurov , I suppose you could use the JavaScript option to leverage Sphinx for your nodejs projects...

blaisep-vgs:
With this last commit, if you `make confluence` then it pushes to our confluence cloud (using creds stored in GH).
if you make `html`, then it builds HTML locally.

Issue #62, opened 2021-03-03
stdlib migration strategy


mjallday:
We have a set of common functionality we need to implement in order to have parity with our customers in production today that we must implement.

## What to migrate 

Most of these immediate requirements are already codified in the FCO framework here - https://github.com/verygoodsecurity/fco-catalogue/tree/master/github.com/verygoodsecurity/common/utils

Initially, we do not need to implement the entire library, only the functionality used by customers. However, we should still implement the entire interface and leave the methods unimplemented so we have a clear view of what needs to be implemented later. This can be done by adding empty methods that raise an exception when called. 

## How to migrate

Each lib must be done as a separate pull-request. It must have tests implemented before any logic is migrated.

We need to implement the larky equivalent of these methods. Our approach is to implement the same method signature that the python stdlib provides. For example, for base64 we would implement `base64.b64encode(s, altchars=None)` and it should behave the same way that the python3 base64 library does.

In order to achieve this outcome we will need to assert that our functionality is identical. We can do this by lifting the test suite from the cpython implementation of the library e.g. for base64 we would implement the fixtures and assertions from https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/test/test_base64.py

### Migrate the interface _first_

For this purpose we have created https://github.com/mahmoudimus/py2star which has [a tool which will take a python file and convert it into an empty starlarky file](https://github.com/mahmoudimus/py2star/blob/main/src/py2star/tokenize_signature.py) which can serve as the base for the migration

Given the [cpython implementation of base64](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/base64.py) as an input it should output something like https://github.com/verygoodsecurity/starlarky/blob/22bf9fdcf70cc3b6a7d1ce91e405280f7a67c494/larky/src/main/resources/stdlib/base64.star

This file should provide the interface and allow us to implement a larky native version of the required functionality. We should not change the interface.

### Migrate tests _second_

The tests will need to be rewritten from the cpython version but the asserts should not change. E.g. `self.assertTypedEqual` should become `asserts.assert_that`

We should find the test file in cpython and migrate the asserts across. This may require a small tool similar to `tokenize_signature.py`.

### Implement the logic _third_

[...]

PR #66, opened 2021-03-04
Base64

mjallday:
- Example of test library to implement - https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/test/test_base64.py
- Example of existing larky test library - https://github.com/verygoodsecurity/starlarky/blob/ad603b25094f1b02dda0c557c291b82de5510464/larky/src/test/resources/stdlib_tests/test_re.star

If our test suite is implemented correctly we should see this test failing because there is no logic implemented

Part of #62


mahmoudimus:
Closing in favor of #102 with this commit: https://github.com/verygoodsecurity/starlarky/pull/102/commits/e181789bdab20e69a33b9a0132e3089863528206

PR #113, opened 2021-05-12
[DONOTMERGE] byteliterals + escape sequence for starlark!

mahmoudimus:


larky-api/pom.xml
```diff
@@ -45,8 +45,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>${lombok.version}</version>
-            <scope>compile</scope>
```
jentfoo, line 49: I assume you still want the `compile` scope?  Otherwise this will be pulled in as a transitive dependency

jentfoo: My mistake, this should be `provided` as changed below

mahmoudimus:
Closing in favor of https://github.com/verygoodsecurity/starlarky/pull/115

PR #115, opened 2021-05-16
[✅] ready for merge: starlark: add 'bytes' data type, for binary strings

mahmoudimus:

This change defines a 'bytes' data type, an immutable string of bytes. In this Java implementation of Starlark, ordinary strings use strings of UTF-16 codes for the ordinary string type, and thus need a distinct type for byte strings.

## Documentation

- https://github.com/bazelbuild/starlark/blob/master/spec.md#bytes
- https://github.com/bazelbuild/starlark/blob/master/spec.md#bytes-literals
- https://github.com/bazelbuild/starlark/blob/master/spec.md#hash

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No

Attempts to fix https://github.com/bazelbuild/starlark/issues/112

### DIFF
--- a/libstarlark/src/main/java/net/starlark/java/eval/Eval.java
+++ b/libstarlark/src/main/java/net/starlark/java/eval/Eval.java
@@ -22,10 +22,12 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+
 import net.starlark.java.spelling.SpellChecker;
 import net.starlark.java.syntax.Argument;
 import net.starlark.java.syntax.AssignmentStatement;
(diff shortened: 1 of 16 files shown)


mahmoudimus:
@Agitolyev @aslepakurov I have minimized the changes to the *absolute minimum* to pass the specification so that we can contribute this upstream to https://github.com/bazelbuild/bazel. After reviewing this, feel free to contribute the patch upstream and to have a review from the Starlark team. You can see the relevant folks on this issue: https://github.com/bazelbuild/bazel/issues/374.

aslepakurov (approved):

PR #117, opened 2021-05-18
jsonpath

mahmoudimus:


OmegaPointZero (approved):

la-luo:
So looks like recursive call is disallowed in larky, then can I get some idea how to implement json path with * in expression, if java is discouraged?

mahmoudimus:
@la-luo this is where we have to use a non-recursive descent parser which is just an `LL(1) parser`. I have some ideas on how to solve this but the key here is that we should not be using Java here (*especially*) since we must operate with `jsonpath` with a stream processing.

aslepakurov (approved):

aslepakurov (approved):

PR #118, opened 2021-05-18
standardize the iso8583 package

mahmoudimus:

Follows the `vendor` package standard we have for every other external third-party port.

- [X] Yes
- [ ] No

Instead of importing from `load("@vendor//ISO8583Encoder"...)`, we need to make sure we import from `load("@vendor//iso8583/encoder"...)` etc.

This follows the appropriate library structure for the library we have ported: https://github.com/knovichikhin/pyiso8583/tree/master/iso8583

image

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No


ozoz03 (approved):
looks better indeed

Agitolyev (approved):

PR #124, opened 2021-05-26
Jose

mahmoudimus:
Introduce Python-Jose, XML Parsing, and various stdlib fixes

## NEW 

- [NEW] Introduce try/except/else/finally + tests
- [NEW] Introduce an python-compatible enum module for state transitions.
- [NEW] Introduce stringio, a python compatible string module for Common string operation - https://docs.python.org/3/library/string.html
- [NEW] Introduce operator.length_hint which mimics the len() operator protocol
- [NEW] Introduce an equivalent python `chr` built-in
- [NEW] Introduce the six module for easier library porting.
- [NEW] introduce a itertools port:
  - I couldn't get `itertools.chain.from_iterable` to work as well as `itertools.chain` so I had to expose `from_iterable` as a toplevel export
  - Introduce chain/from_iterable to make iterating over chained iterables easier

## FIXES

- [FIX] Updated semantics (`__name__` instead of `__class__`). Using `__name__` on the class and `__class__` as the type instantiator is the *right* pythonic semantics. Instead of `__class__` being a string, it should be the class factory. `__name__` is the actual name of the "structure" and this change updates the semantics for this. This has a fallback behavior to avoid having to make this a breaking change as well.

- [FIX] Result module updates:
  - [FIX] For errors, we seem to be suppressing stacktraces of where the actual error happened if there is a nested error evaluation. This is an attempt to expose the stacktrace such that when the Error object is unwrapped, the error stack trace is preserved.

- [FIX] With the introduction of `StarlarkByte`, we allow `StarlarkByte` to be converted to `LarkyByte` type (temporary hack until we can merge enhancements on `LarkyByte` to `StarlarkByte`)

- [FIX] `request.star` to use `full_url` (which was apparently a bug!)

- [FIX] Regular expression module updates:
  - [FIX] Update re module to match python interface
  - [FIX] expose python compatible [re.Match.span](https://docs.python.org/3/library/re.html#re.Match.span)
  - [FIX] expose python compatible [re.Match.groupdict](https://docs.python.org/3/library/re.html#re.Match.groupdict) to expose named subgroups support.
  - [FIX] expose python compatible [re.Pattern.pattern](https://docs.python.org/3/library/re.html#re.Pattern.pattern)
  - [FIX] a pattern string from which the pattern object was compiled.

- [FIX] `repr()` changes:
  - [FIX] Change `LarkyObject` `__repr__()` to mimic Python's which is just `<class XXXXXX>` (this could introduce a breaking change if there is anything depending on the repr output).
  - [FIX] Use the previous move for `repr()` to expose better error messages for any `LarkyObjects` that have `__class__` fields exposed
  - [FIX] Update repr for `Partial` and `SimpleStruct`.
  - [FIX] It turns out that pylarky passes the *REPR* ???? back and uses regex matching to communicate back and forth. *WHAT?*

- [FIX] Changes to `__dict__()`:
  - [FIX] In python3, `__dict__` is not a function, it is a structField. so, `o.__dict__()` `=>`` o.__dict__`
  - [FIX] `useStarlarkThread=true` and `structField=true` cannot both be set

- [FIX] Update sets module
  - [FIX] Update sets to mimic python's `.union` method which returns an *IMMUTABLE* copy as opposed to `.update()` which is *MUTABLE* and accompanying test.


mahmoudimus:
@aslepakurov @la-luo can you folks review ? this PR is massive, but has a *TON* of fixes and compatibility changes. I can try to break it up but we need to merge this soon to prevent duplicate work being done (possibly at #133)

mahmoudimus:
@aslepakurov once this PR is merged, we can remove the nimbus-jwt/jose java dependency and the security vulnerability associated with it should go away.

la-luo (approved):

aslepakurov (approved):

PR #126, opened 2021-05-27
add url parse module

la-luo:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


larky/pom.xml
```diff
@@ -7,7 +7,7 @@
 
     <groupId>com.verygood.security</groupId>
     <artifactId>larky</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>urlparse01</version>
```
mahmoudimus, line 10: mistake?

runlarky/pom.xml
```diff
@@ -7,7 +7,7 @@
 
     <groupId>com.verygood.security</groupId>
     <artifactId>runlarky</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>urlparse01</version>
```
mahmoudimus, line 10: mistake?

larky/src/main/resources/stdlib/parse.star
```diff
@@ -0,0 +1,161 @@
+# This module defines a standard interface to break Uniform Resource Locator (URL)
+# strings up in components (addressing scheme, network location, path etc.)
+
+# Direct copy from: https://github.com/python/cpython/blob/3.9/Lib/urllib/parse.py
```
mahmoudimus, line 4: We should move this to `stdlib/urllib` instead of having this in `stdlib`

larky/src/main/resources/stdlib/parse.star
```diff
@@ -0,0 +1,161 @@
...
+    splitresult = _urlsplit(url, scheme, allow_fragments)
+    scheme, netloc, url, query, fragment = splitresult['scheme'],splitresult['netloc'],splitresult['path'],splitresult['query'],splitresult['fragment']
+    if scheme in uses_params and ';' in url:
+        url, params = _splitparams(url)
+    else:
+        params = ''
+    result = dicts.add({'scheme': scheme}, {'netloc': netloc}, {'path': url}, {'query': query}, {'params': params}, {'fragment': fragment})
+    # ParseResult(scheme, netloc, url, params, query, fragment)
+    # return _coerce_result(result)
+    return result
```
mahmoudimus, line 43: if you want to match the interface for parseresult, you can do:

```python
def _ParseResult(**kwargs):
    return larky.mutablestruct(__class__="ParseResult", **kwargs)
```

Then you can just do `result.scheme`, etc.

larky/src/main/resources/stdlib/parse.star
```diff
@@ -0,0 +1,161 @@
...
+
+
+def _decode_args(args, encoding=_implicit_encoding,
+                       errors=_implicit_errors):
+    t = ()
+    for x in args:
+        if x:
+            t = t + (x.decode(encoding, errors),)
```
mahmoudimus, line 51: what is decode? maybe we can use codecs module?

```python
load("@stdlib//codecs, "codecs")
...
t = t + (codecs.decode(x, encoding, errors),)
```

larky/src/main/resources/stdlib/parse.star
```diff
@@ -0,0 +1,161 @@
...
+        if wdelim >= 0:                    # if found
+            delim = min(delim, wdelim)     # use earliest delim position
+    return url[start:delim], url[delim:]   # return (domain, rest)
+
+def _noop(obj):
+    return obj
+
+def _encode_result(obj, encoding=_implicit_encoding,
+                        errors=_implicit_errors):
+    return obj.encode(encoding, errors)
```
mahmoudimus, line 69:
```python
load("@stdlib//codecs, "codecs")
return codecs.encode(obj, encoding, errors)
```

mahmoudimus (changes requested):
Great job!

larky/pom.xml
la-luo: yes, this is not ready to be merged, just to share with drew for progress update. I will clean up the code after at the end

larky/src/main/resources/stdlib/parse.star
la-luo: sure. I will check this one

larky/src/main/resources/stdlib/parse.star
la-luo: Thank you! Will update it

larky/src/main/resources/stdlib/parse.star
la-luo: yes, that is my plan, putting it here temporarily for convenience

aslepakurov (approved):

PR #130, opened 2021-06-07
add time module

la-luo:
Python Datetime module relies on Time module which is implemented by C, so I write java module to imitate it. But after exploring the operation pipeline use cases in history which use time related stuff, I found those 3 basic methods in time module already support what our use cases need.

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


larky/pom.xml
```diff
@@ -219,6 +219,11 @@
             <version>4.41.1</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>joda-time</groupId>
```
mahmoudimus, line 223: We do not need this dependency.

larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
```diff
@@ -22,18 +22,7 @@
...
-import com.verygood.security.larky.modules.CryptoModule;
-import com.verygood.security.larky.modules.HashModule;
-import com.verygood.security.larky.modules.JsonModule;
-import com.verygood.security.larky.modules.OpenSSLModule;
-import com.verygood.security.larky.modules.ProtoBufModule;
-import com.verygood.security.larky.modules.RegexModule;
-import com.verygood.security.larky.modules.ResultModule;
-import com.verygood.security.larky.modules.StructModule;
-import com.verygood.security.larky.modules.VaultModule;
+import com.verygood.security.larky.modules.*;
```
mahmoudimus, line 25: Unintentional?

mahmoudimus (changes requested):
Why do we need time related items? Do you mind if you document _exactly_ what required the use of time? I *personally* do not think we should expose timers in Larky as it makes super easy to create timing attacks. Can we review https://developers.cloudflare.com/workers/learning/security-model#step-1-disallow-timers-and-multi-threading and see why we would need time at all?

PR #136, opened 2021-06-25
support Xml attrib, update and serialize

la-luo:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus:
Wow @la-luo - this is phenomenal. I've got a few comments but this rocks!

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -806,7 +806,15 @@ def _ElementTree(element=None, file=None):
                                     of start/end tags
 
         """
-        return Error("Unsupported write method")
+        # return Error("Unsupported write method")
```
mahmoudimus, line 809: Just remove this comment

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -806,7 +806,15 @@ def _ElementTree(element=None, file=None):
...
+        qnames, namespaces = _namespaces(self._root, default_namespace)
+        _serialize_xml(
```
mahmoudimus, line 811: If you expand the original code, you'll see that there's several serialization methods, maybe what we should do is uncomment  lines 818 to 820 and change line 820 from `elif method not in _serialize` to `elif method != 'xml'`, then we return an error. *THEN* we invoke `_serialize_xml()`

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -931,8 +939,14 @@ def _namespaces(elem, default_namespace=None):
                 .build()
 
     # populate qname and namespaces table
-    for elem in elem.iter():
-        tag = elem.tag
+    # for elem in elem.iter():
+    qu = [elem]
```
mahmoudimus, line 943: Great job going BFS instead of recursion. This is a great pattern. Can you document it in https://github.com/mahmoudimus/py2star/blob/main/MIGRATING.md?

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -941,72 +955,68 @@ def _namespaces(elem, default_namespace=None):
...
+                break
+            else:
+                current = qu.pop()
+                new_elems.append(current)
+                qu.extend(current._children)
+    return new_elems
 
 def _serialize_xml(
-    write, elem, qnames, namespaces, short_empty_elements, **kwargs
+    write, root_elem, qnames, namespaces, short_empty_elements, **kwargs
```
mahmoudimus, line 986: when we say "root_elem" do we really mean we can only serialize an entire xml tree, not an individual node?

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -941,72 +955,68 @@ def _namespaces(elem, default_namespace=None):
...
             if types.is_instance(value, QName) and value.text not in qnames:
                 add_qname(value.text)
-        text = elem.text
+        text = current.text
         if types.is_instance(text, QName) and text.text not in qnames:
             add_qname(text.text)
+        qu.extend(current._children)
     return qnames, namespaces
 
+def prepare_elems_for_serialize(tops_level_elems):
```
mahmoudimus, line 971: Wouldn't this function be called `flatten_nested_elements` and introduce a docstring to say that we're building an iterative traversal due to lack of recursion support?

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -941,72 +955,68 @@ def _namespaces(elem, default_namespace=None):
...
+    new_elems = []
+    for el in tops_level_elems:
+        new_elems.append(el)
+        qu = el._children[0:]
```
mahmoudimus, line 975: This is a good catch, so we have to do `list[0:]` to copy the list instead of python `list[:]`, so it is *intentional* to put the 0 before the `:` (colon).

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
```diff
@@ -941,72 +955,68 @@ def _namespaces(elem, default_namespace=None):
...
-        if not elems:
-            break
-        elem, namespaces = elems.pop()
-
+    elems = prepare_elems_for_serialize([root_elem])
+    unclosed_elems = []
+    for elem in elems:
         tag = elem.tag
+        # print('iter elem:', tag)
         text = elem.text
```
mahmoudimus, line 993: This serialization method is ignoring comments and processing instructions and only attributes for now, is that accurate to say?

larky/src/main/resources/stdlib/xmllib.star
```diff
@@ -649,7 +649,7 @@ def XMLParser(**kw):
         if self.__seen_starttag and not self.stack:
             self.syntax_error('multiple elements on top level')
         k, j = tag.span('attrs')
-        attrdict, nsdict, k = self.parse_attributes(tagname, k, j)
+        attrdict, nsdict, k = self.parse_attributes(tagname, i+k, i+j)
```
mahmoudimus, line 652: Can you help me understand why this change was needed?

larky/src/test/resources/stdlib_tests/xml/etree/test_elementtree.star
```diff
@@ -87,12 +88,36 @@ def _test_xpath():
...
+    parser = SimpleXMLTreeBuilder.TreeBuilder()
+    data = ''.join(['<data><teacher name="Jenny"><born>1983</born></teacher>',
+    '<student name="Tim"><performance><Grade>A+</Grade></performance><info><born>2005</born></info></student>',
+    '<student name="John"><performance><Grade>B</Grade></performance><info><born>2004</born></info></student></data>'])
+
+    tree = parse(data, parser)
+    root = tree.getroot()
+
+    # test update node text
+    root.findall(".//*[@name='John']/performance/Grade")[0].text = 'A-'
```
mahmoudimus, line 106: holy shit, this works?

mahmoudimus (changes requested):

larky/src/test/resources/stdlib_tests/xml/etree/test_elementtree.star
la-luo: yes, works on fields like attrib, text, tag in Element mutablestruct

larky/src/main/resources/stdlib/xmllib.star
la-luo: Because k, j are the current tag's attrib indices, i is the current tag index in the whole xml string, if not plus i, it will keep getting attrib between k and j in the first tag, not the current tag

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: It was not finished. I have added support to comments and processing instructions in the newest update

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: updated

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: no, I have showed it in the latest update that it works on child element. I changed it to start_elem, it was for me to differentiate variables as I am also using 'elem' in function inner statement

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: will do

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: Could you check the latest update, I have changed it to be inner function _write just for tostring to use, which does _xml_serialize only?

larky/src/main/resources/stdlib/xml/etree/ElementTree.star
la-luo: I feel like it is better to keep it just in case some user mistakenly uses write method as official doc says. I add a new inner method _write for tostring to use

larky/src/main/resources/stdlib/xmllib.star
mahmoudimus: Ah! Thank you. I think we should put that exact comment above that line so if we ever revisit we don't have to come to the pull request to find out why it's adding i. This is great, thank you!

aslepakurov (approved):

PR #138, opened 2021-06-27
Fix ISO8583 length encoding

vmykh:
Fix ISO8583 length encoding, since encoder uses incorrect padding for fields with variable length

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
```diff
@@ -0,0 +1,968 @@
+load("@stdlib//binascii", unhexlify="unhexlify", hexlify="hexlify")
+load("@stdlib//builtins", "builtins")
+load("@stdlib//unittest", "unittest")
+load("@vendor//asserts", "asserts")
+load("@vendor//iso8583/decoder", Decoder="decoder")
+load("@vendor//iso8583/encoder", Encoder="encoder")
+
+
+def CacaoTestCase_test_decode_encode():
```
mahmoudimus, line 9: What is `Cacao`?

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
```diff
@@ -0,0 +1,968 @@
...
+    payload = bytes(cacao_payload, encoding='ascii')
+    decoded, encoded = Decoder.decode(payload, cacao_specs)
+    print('decoded:' + str(decoded))
```
mahmoudimus, line 12: Are these print statements needed?

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
```diff
@@ -0,0 +1,968 @@
...
+    asserts.assert_that(decoded['2']).is_equal_to('4111111111111111')
+    asserts.assert_that(decoded['14']).is_equal_to('2503')
+    asserts.assert_that(decoded['35']).is_equal_to('4111111111111111=2503')
+    asserts.assert_that(decoded['43']).is_equal_to('GOOGLE *TEMPORARY HOLDg.co/helppay#   GB')
+    asserts.assert_that(decoded['60']).is_equal_to('VISAB017+0000000')
+    asserts.assert_that(decoded['123']).is_equal_to('')
+    asserts.assert_that(decoded['126']).is_equal_to('& 0000600124! Q100002 9 ! Q200002 09! C000026 603  001          70 100 0! C400012 000000003070! 0400020             Y')
+
+    encoded2_raw, encoded2 = Encoder.encode(decoded, cacao_specs)
+    asserts.assert_that(str(encoded2_raw)).is_equal_to(cacao_payload)
```
mahmoudimus, line 28: is there a reason why we have to do `str()` here? what's the original type?

larky/src/main/resources/vendor/iso8583/encoder.star
```diff
@@ -399,16 +399,21 @@ def _encode_field(doc_dec, doc_enc, field_key, spec):
         # BCD LLVAR length \x99 must be string "99"
         # BCD LLLVAR length \x09\x99 must be string "0999"
         # BCD LLLLVAR length \x99\x99 must be string "9999"
+        # "{:0{len_type}d}".format(enc_field_len, len_type=len_type * 2)
+        enc_field_len_str = str(enc_field_len)
```
mahmoudimus, line 403: Is this a bug in the actual upstream library code?

vmykh: nope, upstream library code is correct, but it seems that string formatting wasn't translated properly into larky (original code is commented on line 402, btw)

e.g., for BCD case if `enc_field_len = 5` and `len_type = 2`, original code will result into:
`{:0{len_type}d}".format(enc_field_len, len_type=len_type * 2)` => `"0005"`

but in larky code we had:
`"{enc_field_len}".format(enc_field_len=enc_field_len)` => `5`

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
vmykh: We are currently doing ISO8583 integration for Tribal. Cacao is their payment provider, as far as I understand

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
vmykh: not really, I can remove them

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
vmykh: `type(encoded2_raw) == "bytearray"`

vmykh: but original payload encoding is ascii, not binary, that's why I use strings in this test

larky/src/main/resources/vendor/iso8583/encoder.star
vmykh: `"{enc_field_len}".format(enc_field_len=enc_field_len)` is equivalent to just `str(enc_field_len)`, btw

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
mahmoudimus: Can you verify if it is ascii or latin1 encoding?

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
mahmoudimus: Mind if we rename the tests to something a bit more generic (ie test length encoding) instead of naming vendors or customers in the test?

larky/src/main/resources/vendor/iso8583/encoder.star
mahmoudimus: Ah I see, the padding was the issue. Ok. That makes sense

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
vmykh: I can rename it to `Base24TestCase_test_decode_encode`, for example

according to docs from cacao, the format of ISO8583 is Base24 here (maybe with few differences)

the doc itself is in this ticket:
https://verygoodsecurity.atlassian.net/browse/CSG-116 (attachment: `1. Cacao - Manual ISO8583 v3.1.pdf`)

larky/src/test/resources/vendor_tests/iso8583/test_cacao_payload.star
vmykh: nope, I will need to double check it with the customer

at least, the test payload they sent to us contained only ascii symbols

PR #147, opened 2021-07-30
removed versioning from larky-api

moehajj:
Remove versioning from Larky API. The goal is to focus on backward compatibility as much as possile.

## Risks of this release
User cannot control engine version anymore.

- [x] Yes
- [ ] No

Larky API interface and class name changed.

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


aslepakurov (approved):

larky-api/src/main/java/com/verygood/security/larkyapi/LarkyAPIEngine.java
```diff
@@ -0,0 +1,39 @@
+package com.verygood.security.larkyapi;
+
+import com.verygood.security.larky.jsr223.LarkyScriptEngine;
+import com.verygood.security.larky.parser.ParsedStarFile;
+
+import javax.script.ScriptContext;
+import javax.script.ScriptException;
+import java.io.Reader;
+
+public class LarkyAPIEngine extends LarkyScriptEngine {
```
mahmoudimus, line 10: Does this have to be called `Engine`? Wouldn't this just be a `LarkyRuntime` or something?

larky-api/src/main/java/com/verygood/security/larkyapi/LarkyAPIEngine.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private Object extractStarfileOutput(ParsedStarFile starfile, String outputVar) {
+        return starfile.getGlobalEnvironmentVariable(outputVar, Object.class);
+    }
+
+    public Object executeScript(String script, String outputVar)
+        throws ScriptException, NullPointerException {
```
mahmoudimus, line 17: Also, `NullPointerException` is a runtime error, why would we put it as part of the method parameter name?

PR #149, opened 2021-08-01
Better Struct module

mahmoudimus:

Updated the Struct Module and passed all the relevant tests from https://github.com/python/cpython/blob/b08c48e/Lib/test/test_struct.py

## Documentation

The struct library, https://docs.python.org/3/library/struct.html, is a very pervasive library in the Python eco-system and has lots of usage in the cryptography libraries. The previous implementation was very naive and was barely enough to pass simple tests but as more sophisticated use cases become wide-spread, having it properly tested and implemented allows for bug-free guarantees.

## Risks of this release

None.

- [ ] Yes
- [X] No

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No

This is a 100% re-write of the struct module with 95% code coverage.

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/modules/SysModule.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/SysModule.java
@@ -6,11 +6,19 @@ import net.starlark.java.annot.StarlarkBuiltin;
 import net.starlark.java.annot.StarlarkMethod;
 import net.starlark.java.eval.StarlarkValue;
 
-
+/**
+ * A very simple module that just exposes basic information <b><i>JUST for portability
+ * reasons.</i></b>
+ *
+ * This module will just expose things like <code>maxint</code> or <code>byteorder</code>,
+ * just to make it easier to port certain things.
+ *
+ * There is *NO* intention to making this 100% compatible with Python's sys module.
+ */
 @StarlarkBuiltin(
     name = "sys",
     category = "BUILTIN",
-    doc = "Larky system-specific parameters and functions")
+    doc = "*Larky* system-specific parameters and functions")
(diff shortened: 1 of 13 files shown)


larky/src/main/java/com/verygood/security/larky/modules/SysModule.java
```diff
@@ -0,0 +1,29 @@
+package com.verygood.security.larky.modules;
+
+import java.nio.ByteOrder;
+
+import net.starlark.java.annot.StarlarkBuiltin;
+import net.starlark.java.annot.StarlarkMethod;
+import net.starlark.java.eval.StarlarkValue;
+
+
+@StarlarkBuiltin(
```
aslepakurov, line 10: Please add description for this class. It's purpose is unclear

mahmoudimus: It was in the commit message, but I will put it in the header. Was going to update it with this:

```A very simple module that just exposes basic information for portability
reasons.

This module will just expose things like `maxint` or `byteorder`, just to make
it easier to port certain things.```

Any objections?

aslepakurov (approved):

larky/src/main/java/com/verygood/security/larky/modules/SysModule.java
mahmoudimus: Addressed this here: https://github.com/verygoodsecurity/starlarky/pull/150/commits/fcfa911ec73e04e348051a6b8a1edb06d1c629d4. 

Plz merge https://github.com/verygoodsecurity/starlarky/pull/150 first and this should resolve this conversation.

PR #151, opened 2021-08-03
Fix message signature generation bug

alimariebaker:
Fixes a bug in the``sign`` method.

Specifically, step 1 needs to return the encoded message bytes, but instead was returning a Result object which triggered the following error in [Step 2a (OS2IP)](https://github.com/verygoodsecurity/starlarky/blob/master/larky/src/main/resources/vendor/Crypto/Signature/pkcs1_15.star#L84): 

`Error in int_from_bytes: in call to int_from_bytes(), parameter 'bytes' got value of type '_JResult', want 'bytes'`

This change unwraps the success value (encoded message bytes per ``EMSA-PKCS1-V1_5-ENCODE`` [function](https://github.com/verygoodsecurity/starlarky/blob/ec02f72a2937327ced545ce6d3e72348235957a9/larky/src/main/resources/vendor/Crypto/Signature/pkcs1_15.star#L151), as defined in `PKCS#1 v2.1 (RFC3447, 9.2)`) from the returned Result object. 

## Documentation
``sign`` method implements ``RSASSA-PKCS1-V1_5-SIGN`` as specified in `section 8.2.1 of RFC8017 <https://tools.ietf.org/html/rfc8017#page-36>`).

- [ ] Yes
- [X] No

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/resources/vendor/Crypto/Signature/pkcs1_15.star
+++ b/larky/src/main/resources/vendor/Crypto/Signature/pkcs1_15.star
@@ -80,7 +80,7 @@ def PKCS115_SigScheme(rsa_key):
         k = ceil_div(modBits,8) # Convert from bits to bytes
 
         # Step 1
-        em = _EMSA_PKCS1_V1_5_ENCODE(msg_hash, k)
+        em = _EMSA_PKCS1_V1_5_ENCODE(msg_hash, k).unwrap()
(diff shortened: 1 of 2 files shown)


mahmoudimus:
👍🏼 great pull request :D

mahmoudimus (approved):

PR #154, opened 2021-08-13
enhance: add vgs PII text analysis API module.

Agitolyev:
Introduce VGS/cerebro module that exposes API for intelligent data analysis of PII data on the VGS platform.
    The current implementation provides a text analysis module to find PII data entities. However, the default implementation of the analyzer engine provides only "CREDIT_CARD" entity recognition through the simple regex. Furthermore, through [JavaSPI](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html), any runtime may introduce the custom implementation of TextPIIAnalyzer service, enabling access to any combination of third-party and custom services for text analysis.
   Responsibility of Cerebro module will include not only text. Text is first because it is the most demanded data medium for our users. In the future, the cerebro module will be extended with image/speech/video analysis as well. Apart from searching for PII data, it may enable users to get, for example, text sentiment without possible exposure to PII data inside the text(or speech/image/video).

## Documentation
Import Example:
`load("@vgs//cerebro/text/analyzer", pii_analyzer="pii_analyzer")`

Methods of pii_analyzer:

`pii_analyzer.analyze`
```
      Performs PII entities search, given supported language.
     
      @param text the text to analyze
      @param language two characters for the desired language in ISO_639-1 format. Defaults to 'EN'.
      @param entities List of PII entities that should be looked for in the text. 
                      If the List is empty or null, the analyzer will look for all supported entities. Defaults to EMPTY_LIST.
      @param scoreThreshold A minimum value for which to return an identified entity, defaults to 0.
     
      @return a Result object, that wraps a list of the found PII entities in the text
```

`pii_analyzer.supported_languages`
```
     @return List of supported languages in ISO_639-1 format
```

`pii_analyzer.supported_entities`
```
     Get the list of PII entities analyzer is capable of detecting
    
    @param language Two characters for the desired language in ISO_639-1 format, Defaults to 'EN'.
    
     @return a Result object, that wraps  a list of names of supported PII entities
```

Introducing custom TextPIIAnalyzer service:
From Java SPI docs

> To register your service provider, you create a provider configuration file, which is stored in the META-INF/services directory of the service provider's JAR file. The name of the configuration file is the fully qualified class name of the service provider, in which each component of the name is separated by a period (.), and nested classes are separated by a dollar sign ($).

You'll need to Implement the TextPIIAnalyzer interface in your service and inject it for SPI discovery as described above.

- [ ] Yes
- [x] No

- [ ] Use previous release
- [ ] Use a feature flag
- [x] No


Agitolyev:
Moved to: https://github.com/verygoodsecurity/starlarky/pull/156

PR #168, opened 2021-09-08
Pgp support

la-luo:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


larky/src/main/resources/vendor/OpenPGP/Crypto.star
```diff
@@ -0,0 +1,533 @@
...
+        else:
+            self._message = packet
+        return self
+    self = __init__(packet)
+
+    def get_cipher(algo):
+        def cipher(m, ks, bs):
+                def _cipher(k, iv):
+                    return m.new(k, mode=AES.MODE_CFB,
+                        IV=iv or b'\0'*bs,
```
jentfoo, line 69: Using a predictable or repeated IV with AES CFB is not secure.  It looks to me like we need to enforce a unique IV here

jentfoo: I am semi wrong here, I was originally thinking of CBC.  I just checked and this is is not recommended by NIST, but may be secure.  Still something to carefully consider

mahmoudimus:
Working on this on PR #255

PR #170, opened 2021-09-12
*MERGE LAST* Introduce Multidict  (Stacked PR #1 or 3)

### LINKED ISSUE #159: Port in a library for working with HTTP headers
## Problem or feature statement
There is often important header values that we are required to work with in the proxy operations. 

The starlark `list<tuple>` structure of the current header object `input.headers()` has a narrow scope of what can be done with it before it becomes very painful to work with (via List comprehension, long conditional statements, needing to iterate over same list multiple times).

Specific use case would be wanting to grab a specific (key, value) header pair where you may not know what the actual value is (like when a vgs token is passed in). You want to remove this pair from the header object and do some transformations to the values.

Another weird thing that I've seen others encounter is the case can change for header keys in weird ways for some headers which is another level of complexity. Would be nice to solve for this without needing to case-change in the script. 

One solution using starlark built-ins is to convert `list<tuple>` to `dict` and `dict` back to `list<tuple>`:

```
curl --location --request POST 'https://{VAULT_ID}.sandbox.verygoodproxy.com/post' \ 
--header 'Content-Type: application/json' \
--header 'key1: value1' \
--header 'key2: value2' \
--data-raw '{
        "name": "ali"
  }'
```

```
load("@stdlib//json","json")
def process(input, ctx):
    headerDict = {k: v for k, v in input.headers()}
    key1 = headerDict.pop('key1')
    headerDict['new-key'] = 'larky-is-cool'
    newHeaders = [(k, v) for k, v in headerDict.items()]
    msg = input.set_headers(newHeaders).set_body(json.dumps( {"poppedKey": key1}))
    return msg
```

However, this is insufficient in cases where a multidict is needed instead. 

## Advised solution
Find a library to port in to make it easier to work with HTTP headers (and perhaps other common properties of an http request or response object)

Examples:

- https://multidict.readthedocs.io/en/stable/
- https://werkzeug.palletsprojects.com/en/2.0.x/wrappers/?highlight=request
- (intentionally omitting **requests** library)

## Testing scenarios
- multidict scenario
- grabbing headers and/or keys that have dynamic values
- being able to pop off and then transform a specific header
- transforming headers in place (without removing from object)
- appending new headers
- removing headers
- case sensitivity/insensitivity?

mahmoudimus:

This module builds on the earlier commits to introduce the `multidict` module which fixes #159.

More on https://multidict.readthedocs.io/en/stable/ here.

multidict has four multidict classes: `MultiDict`, `MultiDictProxy`, `CIMultiDict` and `CIMultiDictProxy`.

```python
>>> dct = CIMultiDict(key='val')
>>> 'Key' in dct
True
>>> operator.getitem(dct, 'Key')  # we do not support dct['Key'] yet
'val'
```

The most common use of this class is:

```python
>>> d = MultiDict([('a', 1), ('b', 2), ('a', 3)])
>>> d
<MultiDict ('a': 1, 'b': 2, 'a': 3)>
```

## Documentation
More on https://multidict.readthedocs.io/en/stable/ here.

- [ ] Yes
- [X] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus:
@alimariebaker FYI

mjallday (approved):

PR #192, opened 2021-10-19
[WIP] XML Security / XML Crypto support for Larky

mahmoudimus:
Start of porting xmlsec to Larky. This will let us sign WSSE, WS-SOAP, etc messages with Larky


mahmoudimus:
Using [@ejoffe/spr](https://github.com/ejoffe/sp) to break this PR into a much smaller one.

PR #197, opened 2021-11-16
Fix XML flatten_nested_elements() to put elements in proper order

OmegaPointZero:
Calling the append() function in ElementTree would prepend the element; calling tostring() would write the elements nested properly, but in a backwards order. 

(insert text here)

## Documentation
(insert text here)

## Risks of this release
If customers have been using the XML library to write XML, it has been writing things in the wrong order (although nested correctly). If the order of the elements in the XML being sent matters, this may change that. 

- [ ] Yes
- [X] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus:
@OmegaPointZero great fix :) I actually did this too at #192

aslepakurov (approved):

PR #198, opened 2021-12-06
support zip/unzip

la-luo:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus:
This is not necessary. Zip is already a built-in https://github.com/verygoodsecurity/starlarky/blob/5f8103c22e40ec33c92d5846ec2849eb481a0e2b/libstarlark/src/test/java/net/starlark/java/eval/MethodLibraryTest.java#L673-L684

PR #217, opened 2021-12-08
feat(compat): migrate python3's functools.cmp_to_key to allow us to convert a python2's cmp= keyword argument for sorter functions into the new python3 key= function.

mahmoudimus:
**Stack**:
- #225
- #224
- #223
- #222
- #221
- #220
- #219
- #218
- #217
- #216
- #215
- #214
- #213
- #212
- #211
- #210
- #209
- #208
- #207
- #206
- #205
- #204
- #203
- #202
- #201

PR #212, opened 2021-12-08
feat: Introduce XMLModule.

mahmoudimus:
**Stack**:
- #225
- #224
- #223
- #222
- #221
- #220
- #219
- #218
- #217
- #216
- #215
- #214
- #213
- #212
- #211
- #210
- #209
- #208
- #207
- #206
- #205
- #204
- #203
- #202
- #201

- feat: Expose XMLModule + x509Module to Larky by updating ModuleSupplier.java
- feat: Introduce XML Namespace Context which allows for a mutable namespace dictionary-like structure. This is important to have sane namespace support in python's `xml.etree.ElementTree`. The following diff will use this as a test.

PR #221, opened 2021-12-08
This diff is necessary (before the next one) which will test the updated xmlllib parser to enable better namespace support for ElementTree

mahmoudimus:
**Stack**:
- #225
- #224
- #223
- #222
- #221
- #220
- #219
- #218
- #217
- #216
- #215
- #214
- #213
- #212
- #211
- #210
- #209
- #208
- #207
- #206
- #205
- #204
- #203
- #202
- #201

- feat: Update ElementPath and ElementTree to support parents (which actually deviates from the built-in from Python library, but this is additive, not breaking).
- feat: Enhance ElementPath and ElementTree with support for considerably more complicated xpath.
- feat: Introduce namespace support for ElementTree and preserve ordering for deterministic serialization.
- feat: Update QName support for xpaths.
- chore: Ensure test passes since we know have order preserving node construction (as per hermeticity requirement of Larky)
- feat: All working tests for the python equivalent of its `indent()` function!
- feat: Introduce C14N 2.0 / Canonicalize support from Cpython 3.9+

### DIFF
--- a/larky/src/main/resources/stdlib/xml/etree/ElementPath.star
+++ b/larky/src/main/resources/stdlib/xml/etree/ElementPath.star
@@ -77,7 +77,7 @@ xpath_tokenizer_re = re.compile(r"".join([
 ]))
 
 
-StopIterating = Error("StopIteration")
+StopIterating = StopIteration
 
 
 def generator(seq):
@@ -95,6 +95,7 @@ def generator(seq):
         self._current += 1
         return self._sequence[self._current]
     self.next = next
+    self.__next__ = next
     return self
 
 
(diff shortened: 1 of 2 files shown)


larky/src/main/resources/stdlib/xml/etree/ElementPath.star
```diff
@@ -255,7 +262,7 @@ def prepare_descendant(next, token):
     elif token[0] == "None":
         tag = token[1]
     else:
-        return Error("SyntaxError: invalid descendant")
+        return Error("SyntaxError: invalid descendant").unwrap()
```
aslepakurov, line 265: Unwrapped errors? I need some context

aslepakurov (approved):

PR #228, opened 2021-12-21
GRPC/REPL run for Larky

aslepakurov:
- GRPC evaluator
- Test GRPC client
- Change in invocatino
- Separate local IDE run and Quarkus

## Documentation
-

- [ ] Yes
- [X] No

- [ ] Use previous release
- [ ] Use a feature flag
- [X] No


runlarky/src/main/proto/larky.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package com.verygood.security.mode.grpc;
+option java_outer_classname = "LarkyProto";
+option java_multiple_files = true;
+
+message LarkyProcessRequest {
+  string script = 1;
+  bytes input = 2;
```
moehajj, line 9: the interface will also include string & Starlarky request / response structs soon

runlarky/src/main/proto/larky.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package com.verygood.security.mode.grpc;
+option java_outer_classname = "LarkyProto";
+option java_multiple_files = true;
+
+message LarkyProcessRequest {
```
mahmoudimus, line 7: @moehajj plz review this

moehajj: @aslepakurov I think we should use the `request` & `response` objects here, instead of `bytes`, for the generic proxy flows. It would be best to wait for the PR that introduces the comprehensive versions of these objects, and this way you can avoid a breaking changes that's expected very soon!

aslepakurov: Ok, I will make just a working grpc version without merging, just to build other changes on this. As soon as you would have proto, please, notify, so that I can make appropriate changes

aslepakurov:
How ???

PR #243, opened 2022-01-20
Update VGS request to have `path` and `query_string`

moehajj:


larky/src/test/resources/vgs_tests/http/test_default_request.star
```diff
@@ -104,10 +104,28 @@ def _test_request_get_full_url():
     asserts.assert_that(request.full_url).is_equal_to(url)
 
 
-def _test_request_get_relative_url():
-    relative_url = '/path;parameters?query=argument'
+def _test_request_get_url():
+    url = 'http://netloc/path;parameters?query=argument#fragment'
```
mahmoudimus, line 108: can you test with duplicate query parameters?

moehajj: done https://github.com/verygoodsecurity/starlarky/pull/243/commits/d0c8694d53736d956f867ecb0f525f649cac2baa !

moehajj:
> 

What about it? We are just returning the `request.query_string` as-is right now, not the dict `request.query`. So we do not need to decide how to handle duplicate special cases, they will appear as is in the string

mahmoudimus (approved):

PR #256, opened 2022-04-22
Allows setting of environment variables in ModuleSupplier so that we can control instantiation of builtins.

mahmoudimus:
This allows us to put global variables in environments. This PR is necessary for the types / classes PR that is to follow.

commit-id:e0fdeae8

Allows us to set globals and predeclared bindings from ModuleSupplier to the Larky environment. 

## Documentation

```java
  public static final ImmutableMap<String, Object> CORE_ENVIRONMENT = ImmutableMap.of(
//    "type", LarkyTypeClass.getInstance() // you can put anything here and it will appear as a predeclared binding in the starlark environment
  );
```

## Risks of this release

None.

- [ ] Yes
- [x] No

### DIFF
--- a/larky/pom.xml
+++ b/larky/pom.xml
@@ -44,11 +44,11 @@
         <google.re2j.version>1.6</google.re2j.version>
         <google.truth.version>1.1.3</google.truth.version>
         <javax.xml.bind.jaxb-api.version>2.3.1</javax.xml.bind.jaxb-api.version>
-        <org.bouncycastle.version>1.70</org.bouncycastle.version>
+        <org.bouncycastle.version>1.71</org.bouncycastle.version>
         <org.conscrypt.version>2.5.2</org.conscrypt.version>
         <org.jetbrains.annotations.version>22.0.0</org.jetbrains.annotations.version>
         <org.junit.version>4.13.2</org.junit.version>
-        <org.mockito.version>3.12.4</org.mockito.version>
+        <org.mockito.version>4.4.0</org.mockito.version>
         <org.projectlombok.version>1.18.22</org.projectlombok.version>
         <org.slf4j.version>1.7.32</org.slf4j.version>
         <os-maven-plugin.version>1.7.0</os-maven-plugin.version>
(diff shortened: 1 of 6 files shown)


Agitolyev:
I'll build it locally and run OP tests with this version to verify.

Agitolyev:
I'll build it locally and run OP tests with this version to verify.

larky/pom.xml
```diff
@@ -194,37 +206,20 @@
...
-            <resource>
-                <directory>src/test/java</directory>
-            </resource>
-            <resource>
-                <directory>src/test/resources</directory>
-                <excludes>
-                    <exclude>META-INF/services/*</exclude>
-                </excludes>
-            </resource>
-        </resources>
```
aslepakurov, line 227: Why though?

larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
```diff
@@ -43,22 +43,28 @@
...
   public static final ImmutableSet<Class<?>> CORE_MODULES = ImmutableSet.of(
     LarkyGlobals.class,
-    PythonBuiltins.class,
-    LarkyType.class
+    PythonBuiltins.class
+  );
+
+  public static final ImmutableMap<String, Object> CORE_ENVIRONMENT = ImmutableMap.of(
+//    "object", LarkyPyObject.getInstance(),
+//    "type", LarkyTypeClass.getInstance()
```
aslepakurov, line 64: Why `CORE` is commented and being mentioned further

aslepakurov (changes requested):
Please address comments

larky/pom.xml
mahmoudimus: This is not needed. Maven will automatically use our proper setup, so this isn't needed.

larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
mahmoudimus: This is just temporary until the next few PRs that *do* use CORE environment. It will just be an empty dict for now, so it is basically status quo (the same thing we do currently). That's why this PR exists, so that we are aware that we can now use this `CORE_ENVIRONMENT` to setup values that are to be exposed to the larky runtime environment.

PR #257, opened 2022-04-22
Remove copy and paste from Evaluator

mahmoudimus:

Attempt to stop the copy & paste from Evaluator so that making changes to it going forward is easier. It's still pretty gnarly code that needs to be re-written, but at least this allows us to generally reason about it now.

- [ ] Yes
- [x] No

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/parser/LarkyEvaluator.java
+++ b/larky/src/main/java/com/verygood/security/larky/parser/LarkyEvaluator.java
@@ -66,10 +66,27 @@ public final class LarkyEvaluator {
     this.environment = createEnvironment(larkyScript.getBuiltinModules(), larkyScript.getGlobals());
   }
 
+  /**
+   * The output of a Larky script is an evaluated {@link Module} that contains
+   * various attributes such as {@link Module#getGlobals()},
+   * {@link Module#getPredeclaredBindings()}, as well as other various items.
+   *
+   * Sometimes, when evaluating a Larky script, there is some output that is generated.
+   *
+   * This interface encapsulates an interface that allows the caller to introspect
+   * the result of a Larky script evaluation.
+   */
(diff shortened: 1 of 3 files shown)


larky/src/main/java/com/verygood/security/larky/parser/LarkyEvaluator.java
```diff
@@ -62,14 +66,33 @@ public LarkyEvaluator(LarkyScript larkyScript, Console console) {
     this.environment = createEnvironment(larkyScript.getBuiltinModules(), larkyScript.getGlobals());
   }
 
-  public Module eval(StarFile content)
+
+  @Builder
+  @Data
+  static class EvaluationResult {
```
Agitolyev, line 72: Please, consider defining a `public interface` like:

```
public interface EvaluationResult {
    boolean hasOutput();
    boolean hasModule();
    
    Object getOutput();
    Module getModule();
}

@Builder
@Data
class InternalEvaluationResult implements EvaluationResult {
    private Object output;
    private Module module;

    public boolean hasOutput() {
      return output != null;
    }

    public boolean hasModule() {
      return module != null;
    }
}

public EvaluationResult eval(StarFile content) {
...
}
```

This way we will have an immutable by default interface, and give a client of the API freedom to implement more specific implementation in their code(like if they want to check and hard type output, or do some checks, etc.)

mahmoudimus: Good suggestion.

mahmoudimus: Addressed https://github.com/verygoodsecurity/starlarky/pull/257/commits/0bb92f5a62ba7cc73ff96a94e3a8e127353e208e

PR #265, opened 2022-04-28
Step 7 of introducing types in Larky:

mahmoudimus:
Introduce very simple concrete implementations of PyObject => LarkyPyObject
Introduce very simple concrete implementations of LarkyType => LarkyTypeObject

Delete the previous LarkyType in `<a href="com.verygood.security">com.verygood.security</a>.larky.modules.types` namespace.

---

**Stack**:
- #287
- #286
- #285
- #284
- #283
- #282
- #281
- #280
- #279
- #278
- #277
- #276
- #274
- #273
- #272
- #271
- #270
- #269
- #268
- #267
- #266
- #265 ⮜
- #264
- #263
- #262
- #261
- #260
- #259

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #260, opened 2022-04-28
Step 2 of introducing types into Larky:

mahmoudimus:
Python-like introspection of method names + callstack entries

1. We need to be able to call the annotated method descriptors when retrieving field names. Starlark already supports this but it's package-private. We expose it via StarlarkEvalWrapper.
2. We expose the StarlarkThread debug frames to access variables (again this is package-private), but we need it for the super() implementation.

---

**Stack**:
- #287
- #286
- #285
- #284
- #283
- #282
- #281
- #280
- #279
- #278
- #277
- #276
- #274
- #273
- #272
- #271
- #270
- #269
- #268
- #267
- #266
- #265
- #264
- #263
- #262
- #261
- #260 ⮜
- #259

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #267, opened 2022-04-28
Step 9 of introducing types in Larky:

mahmoudimus:
- Introduce C3 linearization algorithm as per Python to support multiple inheritance! This now allows us to use method resolution orders to traverse attributes and builds for `super()`
- Introduce the concept of a LarkyProvidedTypeClass. This class allows us to create instances of user defined types, which is the beginning of classes and types in Larky. It is important
to note that the `type()` of a `LarkyProvidedTypeClass` is *actually* the type instance itself. This satisfies the same behavior that Python exhibits.

---

**Stack**:
- #287
- #286
- #285
- #284
- #283
- #282
- #281
- #280
- #279
- #278
- #277
- #276
- #274
- #273
- #272
- #271
- #270
- #269
- #268
- #267 ⮜
- #266
- #265
- #264
- #263
- #262
- #261
- #260
- #259


larky/src/main/java/com/verygood/security/larky/objects/mro/C3.java
```diff
@@ -0,0 +1,78 @@
...
+    while (!toMerge.stream().allMatch(List::isEmpty)) {
+      LarkyType goodCandidate = null;
+      outer:
+      for (List<LarkyType> sublist : toMerge) {
+        if (sublist.isEmpty()) continue;
+        LarkyType candidate = sublist.get(0);
+        // is the candidate present in any of the other lists, as any element after 0?
+        for (List<LarkyType> other : toMerge) {
+          if (other.indexOf(candidate) > 0) {
+            continue outer; // reject the candidate
```
aslepakurov, line 60: Maybe let's `stream().forEach()` `toMerge` just to avaid goto?

PR #271, opened 2022-04-28
Step 13 of introducing types in Larky:

mahmoudimus:
We introduce LarkyFunction and LarkyMethod. These are important in Python and are now first class in Larky.

The previous PRs were building to OO features in Larky and now we have reached this crescendo 🎵🙌.

Python’s object-oriented features are built upon a function based environment. Using non-data descriptors, the two are merged seamlessly.

- Functions stored in class dictionaries get turned into methods when invoked. (LarkyFunction)
- Methods only differ from regular functions in that the object instance is prepended to the other arguments. (LarkyMethod)

By convention, the instance is called self but could be called this or any other variable name.

Classes! Now, when we set a function via `types.new_class`, we automatically check to see if there's a StarlarkFunction (just a normal function) and we then wrap it in a LarkyFunction which understands how to bind itself to an instance.

This PR is accompanied by the implementation of the `__getattribute__` Python method, effectively the starting point of the descriptor protocol.

The expression `obj.x` looks up the attribute `x` in the chain of namespaces for `obj`. If the search finds a descriptor outside the instance `__dict__`, its `__get__()` method is invoked according to the precedence rules listed below.

Instance lookup scans through a chain of namespaces:
1. giving data descriptors the highest priority
2. followed by instance variables
3. then, non-data descriptors
4. then class variables,
5. and, lastly `__getattr__()`, if it is provided.

If a descriptor is found for `a.x`, then it is invoked with: `desc.__get__(a, type(a))`.

See more here: <a href="https://docs.python.org/3.10/reference/datamodel.html#object.__getattribute__">https://docs.python.org/3.10/reference/datamodel.html#object.__getattribute__</a>

---

**Stack**:
- #287
- #286
- #285
- #284
- #283
- #282
- #281
- #280
- #279
- #278
- #277
- #276
- #274
- #273
- #272
- #271 ⮜
- #270
- #269
- #268
- #267
- #266
- #265
- #264
- #263
- #262
- #261
- #260
- #259


larky/src/main/resources/stdlib/types.star
```diff
@@ -324,7 +324,8 @@ def prepare_class(name, bases=(), kwds=None):
         meta = kwds.pop('metaclass')
     else:
         if bases:
-            meta = type(bases[0])
+            meta = larky.type_cls(bases[0])
```
aslepakurov, line 327: Should we make changes everywhere?

larky/src/main/java/com/verygood/security/larky/objects/PyObject.java
```diff
@@ -116,7 +116,7 @@ default Object getField(String name, @Nullable StarlarkThread thread) {
     try {
       getattr_ = this.__getattribute__(name, thread);
     } catch (EvalException ex) {
-      getattr_ = null;
+      getattr_ = GetAttribute.dunderGetAttr(this, name, thread, /*throwExc=*/false);
```
aslepakurov, line 119: You use `dunder...` here and `__...` everywhere else

larky/src/main/java/com/verygood/security/larky/objects/LarkyPyObject.java
```diff
@@ -9,6 +9,9 @@
 import com.verygood.security.larky.objects.type.LarkyType;
```
aslepakurov, line 9: There should be a brief to distinguish between this new Larky objects. Some scheme

aslepakurov (approved):

PR #282, opened 2022-05-10
fix: Mimic the semantics in PyCryptodome and translate them to BouncyCastle

mahmoudimus:
In BouncyCastle, if the expected plaintext length is -1
(as opposed to the default in PyCryptodome of 0)
then it means we do not have an expected plain text length.

We add a patch before we invoke `JCrypto` implementation of PKCS1_v1_5#decode.
Since, BouncyCastle assumes that the expected plaintext is actually 0, as
opposed to what PyCryptodome intends, we mimic the behavior.

---

**Stack**:
- #287
- #286
- #285
- #284
- #283
- #282 ⮜
- #281
- #280
- #279
- #278
- #277
- #276
- #274
- #273
- #272
- #271
- #270
- #269
- #268
- #267
- #266
- #265
- #264
- #263
- #262
- #261
- #260
- #259


larky/src/main/java/com/verygood/security/larky/modules/crypto/Cipher/LarkyPKCS1.java
```diff
@@ -44,46 +45,87 @@ public class LarkyPKCS1 implements StarlarkValue {
...
     if (len_sentinel > len_em_output) {
       return FAIL;
     }
-    if (expected_pt_len > 0 && expected_pt_len > (len_em_output - EM_PREFIX_LEN - 1)) {
-      return FAIL;
+    if (expected_pt_len != -1) {
+      if (expected_pt_len > (len_em_output - EM_PREFIX_LEN - 1)) {
+        return FAIL;
+      }
     }
```
aslepakurov, line 64:
```suggestion
    if (expected_pt_len != -1 || expected_pt_len > (len_em_output - EM_PREFIX_LEN - 1)) {
        return FAIL;
    }
```

aslepakurov (approved):

PR #291, opened 2022-05-19
Fixed JWS Signing with RSA keys

mottersheadt:
Prior to this change, Larky would throw an error when attempting to sign data using a JWK constructed from an RSA public key.

The issue happened because we were calling the pkcs1_signer as a function inappropriately. Fixed issue and added test cases to ensure RSA and ECC do not throw failures. Currently there is no assertion on the output because the output of the encoded_signature value is random every time. 

## Risks of this release
Need to ensure that any other places that use this code are not affected. Currently all the tests are passing.

- [ ] Yes
- [ x] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ x] No


mahmoudimus (approved):
Good stuff

PR #304, opened 2022-08-04
CORE-1464 - Vault alias decorator

k-sever:
Add support of alias decorator in larky vault module

## Documentation
(insert text here)

- [ ] Yes
- [X] No

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/pom.xml
+++ b/larky/pom.xml
@@ -43,7 +43,6 @@
         <google.jimfs.version>1.2</google.jimfs.version>
         <google.re2j.version>1.6</google.re2j.version>
         <google.truth.version>1.1.3</google.truth.version>
-        <google.gson.version>2.9.1</google.gson.version>
         <javax.xml.bind.jaxb-api.version>2.3.1</javax.xml.bind.jaxb-api.version>
         <org.bouncycastle.version>1.71</org.bouncycastle.version>
         <org.conscrypt.version>2.5.2</org.conscrypt.version>
@@ -101,12 +100,6 @@
             <groupId>com.google.flogger</groupId>
             <artifactId>flogger-slf4j-backend</artifactId>
             <version>${google.flogger.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>com.google.code.gson</groupId>
-            <artifactId>gson</artifactId>
-            <version>${google.gson.version}</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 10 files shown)


larky/pom.xml
```diff
@@ -43,6 +43,7 @@
         <google.jimfs.version>1.2</google.jimfs.version>
         <google.re2j.version>1.6</google.re2j.version>
         <google.truth.version>1.1.3</google.truth.version>
+        <google.gson.version>2.9.1</google.gson.version>
```
mahmoudimus, line 46: Why do you need this? We already have a json parser that we've built as part of internal starlark.

larky/src/main/java/com/verygood/security/larky/modules/vgs/vault/spi/LarkyVault.java
```diff
@@ -7,7 +7,7 @@
 
 public interface LarkyVault extends StarlarkValue {
 
-    Object redact(Object value, Object storage, Object format, List<Object> tags) throws EvalException;
+    Object redact(Object value, Object storage, Object format, List<Object> tags, Object decoratorConfig) throws EvalException;
```
mahmoudimus, line 10: Should we move this to a builder pattern instead? Or do we just want to continue expanding the parameters? Does "decoratorConfig" really belong as a parameter ?

larky/pom.xml
mahmoudimus: I am not OK w/ adding another dependency tbh.

larky/src/main/java/com/verygood/security/larky/modules/vgs/vault/spi/LarkyVault.java
Agitolyev: I believe it makes API simpler to have a decorator as a parameter.
1. Builder adds some notion of state inside a script, complicating things.
2. You don't need to instantiate different objects if you have to tokenize different PII entities using different approaches:
``` python
load('@stdlib/json', 'json')
load('@stdlib//builtins', 'builtins')
load('@vgs//vault', 'vault')

def process(input, ctx):
    json_body = json.loads(str(input))
    
    json_body['card_number'] = vault.redact(json_body['card_number'], decoratorConfig=...)
    json_body['ssn'] = vault.redact(json_body['ssn'])
    return json.dumps(json_body)
```
3. Explicit is better than implicit.

larky/pom.xml
k-sever: agree, I didn't see we have one. Will use the internal one, thx

nestorsokil (approved):

Agitolyev (approved):

PR #306, opened 2022-08-15
Expose AES CTR

OmegaPointZero:
* Exposes AES-CTR for use 
* ~Add builtin `list.reverse()` function that was missing~

## Documentation
https://pycryptodome.readthedocs.io/en/latest/src/cipher/classic.html#ctr-mode

- [ ] Yes
- [X] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/libstarlark/src/main/java/net/starlark/java/eval/StarlarkList.java
+++ b/libstarlark/src/main/java/net/starlark/java/eval/StarlarkList.java
@@ -455,17 +455,6 @@ public final class StarlarkList<E> extends AbstractList<E>
     addElement((E) item); // unchecked
   }
 
-  @StarlarkMethod(name = "reverse", doc = "Reverses the order of the list items.")
-  public void reverse() throws EvalException {
-    Object[] revElems = new Object[size];
-    int j = size;
-    for (int i = 0 ; i < size; i++) {
-      revElems[j-1] = elems[i];
-      j = j - 1;
-    }
-    elems = revElems;
-  }
-
   @StarlarkMethod(name = "clear", doc = "Removes all the elements of the list.")
   public void clearElements() throws EvalException {
     Starlark.checkMutable(this);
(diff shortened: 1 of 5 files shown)


libstarlark/src/main/java/net/starlark/java/eval/StarlarkList.java
```diff
@@ -455,6 +455,17 @@ public void append(Object item) throws EvalException {
     addElement((E) item); // unchecked
   }
 
+  @StarlarkMethod(name = "reverse", doc = "Reverses the order of the list items.")
```
mahmoudimus, line 458: We shouldn't modify the StarlarkList file. You can use reversed(list) instead.

mahmoudimus (changes requested):

libstarlark/src/main/java/net/starlark/java/eval/StarlarkList.java
OmegaPointZero: I thought part of our mission with Larky is to make it work as close to python as possible. Is there a reason we shouldn't have this built in functionality? 

I can fix the code, I'm just wondering why we're not gonna implement something built-in to python.

mahmoudimus: StarlarkList is owned upstream by Google's bazel team. We don't want to maintain the differences. 

The only reason it diverges slightly is because we added bytes support and that was not added in upstream bazel, but we should absolutely merge it upstream.

StarlarkList modifications should go to bazelbuild/starlark first and if approved, we can add it. Otherwise, we shouldn't add it.

Also, I don't see the difference between reversed(list) vs list.reverse. Both are valid python.

OmegaPointZero: Ah, that makes perfect sense. Thank you for the explanation.

PR #309, opened 2022-08-16
Expose AES CTR

OmegaPointZero:
## Fixes #306 
Signed Commits

(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/test/resources/quick_tests/test_CTR.star
+++ b/larky/src/test/resources/quick_tests/test_CTR.star
@@ -7,6 +7,20 @@ load("@stdlib//random", "random")
 load("@vendor//Crypto/Cipher/AES", AES="AES")
 load("@vendor//Crypto/Util/Counter", Counter="Counter")
 load("@vendor//asserts","asserts")
+load("@vendor//Crypto/Hash", SHAKE128="SHAKE128")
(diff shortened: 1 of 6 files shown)


larky/src/test/resources/quick_tests/test_CTR.star
```diff
@@ -0,0 +1,33 @@
+load("@stdlib//unittest","unittest")
```
mahmoudimus, line 1: You should absolutely port this test https://github.com/Legrandin/pycryptodome/blob/master/lib/Crypto/SelfTest/Cipher/test_CTR.py to ensure 100% compatibility with pycryptodome. The quick_tests are OK to test something quickly, but for crypto items we should port the pycryptodome tests as we've done previously. See the current test suite here for ciphers: https://github.com/verygoodsecurity/starlarky/tree/master/larky/src/test/resources/vendor_tests/Crypto/Cipher or https://github.com/verygoodsecurity/starlarky/tree/master/larky/src/test/resources/vendor_tests/Crypto for the same test suites that was ported from pycryptodome.

mahmoudimus (approved):
Excellent job!

PR #310, opened 2022-08-18
Alias decorator config moved out from modules root

k-sever:
Follow up to https://github.com/verygoodsecurity/starlarky/pull/304
Moved Alias decorator config out from modules root

## Documentation
(insert text here)

- [ ] Yes
- [X] No

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus (approved):
Thank you. Excellent!

PR #311, opened 2022-08-19
Fix incorrect mac generation when using a CBC Block Cipher via Buffered Block Cipher.

mahmoudimus:

fix(bug): When using a BufferedBlockCipher, check if there's any further buffer left over before calling doFinal.

In BouncyCastle, we should be using `doFinal` without worrying about
the buffer size or if there's any partial buffer left over to process.
However, when the logic is split between pycryptodome (specifically
for the padding logic) and BouncyCastle, we have to add this check in
to make sure we're not issuing a `reset()` on the internal state of
the block cipher.

In addition, if we were to re-use the cipher (as pycryptodome does if
the caller passes in `msg_len` or `assoc_len`), then the `doFinal`
state can cause incorrect output.

Without this diff that adds a re-entrant init() check as well as a
check to see if there's any more data to process, we will not be able
to verify the MAC on any AEAD cipher that uses a legacy mode cipher
since we will not invoke `doFinal()`. BouncyCastle will (correctly)
attempt to calculate the MAC, however, this part is done in
pycryptodome and as a result, we need to add the check to see if there
is more data left to process.

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/test/resources/quick_tests/test_justin.star
+++ b/larky/src/test/resources/quick_tests/test_justin.star
@@ -41,17 +41,17 @@ def _sign_header_and_claims(encoded_header, encoded_claims, algorithm, key):
     k = jwk.construct(key, algorithm)
     encoded_signature = base64url_encode(k.sign(signing_input))
     encoded_string = b".".join([encoded_header, encoded_claims, encoded_signature])
-    print(encoded_string)
+    # print(encoded_string)
     return encoded_string
 
 
 def sign(payload, key, headers=None, algorithm=ALGORITHMS.HS256):
     encoded_header = _encode_header(algorithm, additional_headers=headers)
     encoded_payload = _encode_payload(payload)
-    print("======================")
-    print(encoded_header)
-    print(encoded_payload)
-    print("======================")
+    # print("======================")
+    # print(encoded_header)
+    # print(encoded_payload)
+    # print("======================")
     signed_output = _sign_header_and_claims(encoded_header, encoded_payload, algorithm, key)
     return signed_output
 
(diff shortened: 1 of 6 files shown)


larky/src/test/resources/vendor_tests/Crypto/Cipher/test_CBC.star
```diff
@@ -0,0 +1,572 @@
+# # ===================================================================
```
mahmoudimus, line 1: Will fill this in later.

larky/src/test/resources/quick_tests/test_justin.star
```diff
@@ -41,17 +41,17 @@ def _sign_header_and_claims(encoded_header, encoded_claims, algorithm, key):
...
 
 
 def sign(payload, key, headers=None, algorithm=ALGORITHMS.HS256):
     encoded_header = _encode_header(algorithm, additional_headers=headers)
     encoded_payload = _encode_payload(payload)
-    print("======================")
-    print(encoded_header)
-    print(encoded_payload)
-    print("======================")
+    # print("======================")
```
Agitolyev, line 51: Are those debug leftovers?

Agitolyev (approved):

larky/src/test/resources/quick_tests/test_justin.star
mahmoudimus: Yeah, I just wanted to stop the print output during test execution. I didn't write this `test_justin.star` but I don't want test output polluted by `print()` statements.

PR #317, opened 2022-09-20
Fix Jose JWE with certificates

OmegaPointZero:
* Fixes bug in python-jose that prevented certificates from being used in place of public PEM keys
* Adds python-jose test to cover the use case
* Adds `test_customer_usecase.star` under `quickTests` to provide a template for debugging Larky using the same format and methods as `process(input, ctx)` within the YAML

There was a block of code in the `pycrypto_backend.star` that was expecting to have to parse a certificate differently than a normal PEM file, and did not have any corresponding method to load a certificate. Removing these lines allows jose to behave the way it does in python.

## Documentation
(insert text here)

- [ ] Yes
- [X] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


mahmoudimus (approved):
Wow @OmegaPointZero - great job mate!

PR #320, opened 2023-02-02
Add ECDH

OmegaPointZero:
Pycryptodome has been the backbone of Starlarky's encryption capabilities. However, Pycryptodome is missing ECDH capabilities (despite having ECDSA). The other cryptographic modules that would have ECDH use pycryptodome as the backend, and are thus missing this capability as well. This ports it in a standalone way, so that if pycryptodome adds it in the future, there is no need to worry about documenting additional differences between our libraries. 

## Documentation
See `test_ecdh.star` for usage- will write wiki page soon. 

- [ ] Yes
- [X] No

- [X] Use previous release
- [ ] Use a feature flag
- [ ] No

* This was implemented in this standalone way rather than using the pycryptodome EC Libraries due to difficulties in getting the point-jacobi multiplication to work in Larky without recursion or mutating frozen dicts. 
* This actually allows us to do the decryption only on Apple Pay tokens completely within Larky right now- there is still additional work to be done to extract information from the certificate they supply and to perform all of the validations, but this lets us do the decryption itself. 
* This is a prerequisite to porting [applepay](https://github.com/halturin/applepay). EC Keys from this library's test case were included in our test files to ensure sufficient EC Key capabilities were ported.

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
+++ b/larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
@@ -27,6 +27,7 @@ import com.verygood.security.larky.modules.ChaseModule;
 import com.verygood.security.larky.modules.CodecsModule;
 import com.verygood.security.larky.modules.CollectionsModule;
 import com.verygood.security.larky.modules.CryptoModule;
+import com.verygood.security.larky.modules.ECDHModule;
 import com.verygood.security.larky.modules.JsonModule;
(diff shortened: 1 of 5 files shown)


SMony-L (approved):
LGTM

charliefoster-vgs (approved):
Looks good.

Lexora (approved):

alejohern (approved):

Iapetus999 (approved):

PR #326, opened 2023-03-14
WIP [PAYOP-1116] Implement Larky function for rendering network token

fangpenlin:
## Description of changes in release / Impact of release:
Implement Larky function for rendering network token

## Documentation
Created a new network token module with service interface to be implemented later.

## Risks of this release
The default module service either NoopNetworkTokenService or MockNetworkTokenService basically does nothing so shouldn't introduce any risk.

### Is this a breaking change?
- [ ] Yes
- [x] No

### Is there a way to disable the change?
- [ ] Use previous release
- [ ] Use a feature flag
- [x] No

PR #327, opened 2023-03-14
[PAYOP-1116] Implement Larky function for rendering network token (ALT implementation)

fangpenlin:
Implement Larky function for rendering network token

## Documentation
Created a new network token module with service interface to be implemented later.

## Risks of this release
The default module service either NoopNetworkTokenService or MockNetworkTokenService basically does nothing so shouldn't introduce any risk.

This alt implementation get rid of json-path deps in #326, the attack vector is now greatly reduced.

- [ ] Yes
- [x] No

- [ ] Use previous release
- [ ] Use a feature flag
- [x] No

[PAYOP-1116]: https://verygoodsecurity.atlassian.net/browse/PAYOP-1116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


larky/src/main/resources/vendor/vgs/nts_helpers.star
```diff
@@ -0,0 +1,25 @@
+load("@stdlib//larky", larky="larky")
+load("@vendor//jsonpath_ng", jsonpath_ng="jsonpath_ng")
+load("@vgs//nts", "nts")
+
+
+def render(input, pan, exp_month=None, exp_year=None, cryptogram_value=None, cryptogram_eci=None):
```
mottersheadt, line 6: I like this a lot. This way, in more complex (non-JSON) integrations, we can just call "nts.get_netwok_token(pan)" to get the exact data and fill in the fields as needed.

PR #334, opened 2023-05-04
Payop 1291 1292

jitendra-xoriant:
https://verygoodsecurity.atlassian.net/browse/PAYOP-1291
https://verygoodsecurity.atlassian.net/browse/PAYOP-1292

(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/resources/vgs/nts.star
+++ b/larky/src/main/resources/vgs/nts.star
@@ -1,4 +1,5 @@
 load("@vgs//native_nts", _nts="native_nts")
+load("@vgs//vault", vault="vault")
 load("@stdlib//larky", larky="larky")
 load("@vendor//jsonpath_ng", jsonpath_ng="jsonpath_ng")
 
@@ -9,6 +10,7 @@ def render(
     cvv,
     amount,
     currency_code,
+    dcvv=None,
     exp_month=None,
     exp_year=None,
     cryptogram_value=None,
@@ -35,6 +37,8 @@ def render(
            rendered and injected into the payload.
     :param cvv: JSONPath to the CVV of the credit card in the input payload. Used to pass to the network for retrieving
            the corresponding network token and cryptogram to be returned.
+    :param dcvv: JSONPath to the DCVV of the credit card in the input payload. Used to pass to the network for retrieving
+               the corresponding network token and cryptogram to be returned.
     :param amount: JSONPath to the amount of payment for the transaction to be made with the network token in the input
            payload. Used to pass to the network for retrieving the corresponding network token and cryptogram to be
            returned.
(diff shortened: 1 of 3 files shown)


larky/src/main/resources/vgs/nts.star
```diff
@@ -73,7 +79,12 @@ def render(
     return input
 
 
+def supports_dcvv(input, pan):
+    return vault.getValue(jsonpath_ng.parse(pan)).startswith(4)
```
viacheslav-fomin-main, line 83: we should not use vault here, we must pass a preprocessed value here

mjallday: how do we work around this slava? without this we'd need to put the logic into the main script which is less ideal (we could if we must but the encapsulation will look wrong). could we pass a method reference?

e.g.

```python
nts.supports_dcvv(payload, selector, resolver=lambda: vault.get)
```

larky/src/main/resources/vgs/nts.star
```diff
@@ -1,18 +1,20 @@
...
+        input,
+        pan,
+        cvv,
+        amount,
+        currency_code,
+        dcvv=None,
+        exp_month=None,
+        exp_year=None,
+        cryptogram_value=None,
+        cryptogram_eci=None,
```
fangpenlin, line 17: Code style. We should use the same indentation in the code base, or at least for the same file. 
A good way to find out what's changed intentionally or not is to review your own code.
Auto code formatter might also help.

larky/src/main/resources/vgs/nts.star
```diff
@@ -50,12 +54,14 @@ def render(
     """
     pan_value = jsonpath_ng.parse(pan).find(input).value
     cvv_value = jsonpath_ng.parse(cvv).find(input).value
+    dcvv_value = jsonpath_ng.parse(dcvv).find(input).value
```
fangpenlin, line 57: this value suppose to be optional? I guess `cvv` and `dvcc`, in real life we probably is going to use one of them but not both of them, right? @mjallday

if so, we may need to use only one of them here?

larky/src/main/resources/vgs/nts.star
fangpenlin, line 83: it seems like no unit test for this new method added here in the PR?

larky/src/main/resources/vgs/nts.star
mjallday: yes, it's an either / or. no _and_

larky/src/main/resources/vgs/nts.star
mjallday: please yes! run it as a ci check to make sure everything's linted 🙏 🙌

PR #341, opened 2023-06-06
[PAYOP-1372] Fix json path bugs

fangpenlin:
Fix json path bugs

## Documentation
Fix json path bugs

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

[PAYOP-1372]: https://verygoodsecurity.atlassian.net/browse/PAYOP-1372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/larky/src/main/resources/vendor/jsonpath_ng.star
+++ b/larky/src/main/resources/vendor/jsonpath_ng.star
@@ -30,9 +30,10 @@ def _write(node, keys, value):
     data = node
     for index in range(1, len(keys)):
         key = keys[index]
+        is_last_key = index == len(keys) - 1
         if types.is_dict(data):
             # If this is the leaf node, write
-            if index == len(keys) - 1:
+            if is_last_key:
                 # print('data to update:', data)
                 data[key] = value
                 return node
@@ -44,7 +45,12 @@ def _write(node, keys, value):
                 data = data[key]
 
         elif types.is_list(data):
-            if key > -1 and key < len(data):
+            # If this is the leaf node, write
+            if is_last_key:
+                # print('data to update:', data)
+                data[key] = value
+                return node
+            elif key > -1 and key < len(data):
                 data = data[key]
             else:
                 fail("ParsingException('Key \"{%s}\" does not exist in node')" % key)
--- a/larky/src/main/resources/vendor/jsonpath_ng.star
+++ b/larky/src/main/resources/vendor/jsonpath_ng.star
@@ -5,7 +5,6 @@ load("@stdlib//types", types="types")
 
 _WHILE_LOOP_EMULATION_ITERATION = larky.WHILE_LOOP_EMULATION_ITERATION
 
-
 def _read(node, keys):
     data = node
     # First key should always be $, ignore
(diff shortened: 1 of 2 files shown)


larky/src/test/resources/vendor_tests/test_jsonpath.star
```diff
@@ -52,6 +56,43 @@ def _test_simple_jsonpath():
...
+        expr = jsonpath_ng.parse(path)
+        result = expr.find(FIXTURE)
+        asserts.assert_that(result.value).is_equal_to(expected)
+
+
+def _test_get_with_quoted_key():
+    data = {
+        "card[number]": "4242424242424242",
+        "card[cvc]": "123",
+        "card[expire_year]": ["2024"],
```
Iapetus999, line 73: why would someone do this??

fangpenlin: an example of larky script looks like this

```starlarky 
load("@stdlib//builtins", "builtins")
load("@stdlib//json", "json")
load("@stdlib//urllib/parse", "parse")
load("@vendor//option/result", safe="safe")
load("@vgs//nts", "nts")
load("@vgs//vault", vault="vault")

# Wrap the network token render function in the safe decorator to capture
# network token errors so that it will return a result object instead of
# failing the whole request.
safe_render = safe(nts.render)

def process(input, ctx):
  # Parse form-encoded body data of request
  body = parse.parse_qs(str(input.body))
  result = None
  if (
    # Check and see if use network token header enabled
    nts.use_network_token(input.headers)
  ):
    # Render network token into the body payload
    result = safe_render(
      body,
      pan="$['payment_method_data[card][network_token][number]'][0]",
      cvv="$['payment_method_data[card][cvc]'][0]",
      amount="$.amount[0]",
      currency_code="$.currency[0]",
      exp_month="$['payment_method_data[card][network_token][exp_month]'][0]",
      exp_year="$['payment_method_data[card][network_token][exp_year]'][0]",
      cryptogram_value="$['payment_method_options[card][network_token][cryptogram]']",
      cryptogram_eci="$['payment_method_options[card][network_token][electronic_commerce_indicator]']",
    )
  if result != None and result.is_ok:
    # We got the network token and ran render successfully, let's use it here
    body = result.unwrap()

  else:
    # Network token is not available, let's fallback and reveal the raw card number
    body["payment_method_data[card][number]"] = [vault.reveal(body.pop("payment_method_data[card][network_token][number]")[0])]

    # Reveal CVC value if available
    cvv = body.get("payment_method_data[card][cvc]", [None])[0]
    if cvv != None:
      body["payment_method_data[card][cvc]"] = [vault.reveal(cvv, storage="volatile")]

    # Also remove unused fields meant for network token
    body.pop("payment_method_data[card][last4]")
    body.pop("payment_method_data[card][network_token][exp_month]")
    body.pop("payment_method_data[card][network_token][exp_year]")

  if body.get("payment_method_data[card][cvc]", [""])[0].startswith("tok_"):
    # The cvc token probably expired already, so reval doesn't change
    # the value, let's remove it to avoid PSP returning error
    body.pop("payment_method_data[card][cvc]")

  # Convert body object back to form-encoded data
  input.body = builtins.bytes(parse.urlencode(body, doseq=True))
  return input
```

Iapetus999: I think I understand the bug. If the key contains a `]` it sometimes got interpreted as the _end_ of the key instead of being _part_ of the key 💡

larky/src/main/resources/vendor/jsonpath_ng.star
```diff
@@ -113,9 +131,9 @@ def _parse(query):
                 keys.append(element)
                 element = ""
 
-            close = query.find("]", index)
```
Iapetus999, line 116: This line is the bug

Iapetus999 (approved):

PR #398, opened 2023-10-06
[VAULT-2589][Larky Logging]Integrate Starlark logging with Horizon

shreyoshi-mahato:
added ConsoleScriptContext new eval method override


Iapetus999:
I don't understand why we need this, I thought we could do all this in Horizon

mahmoudimus:
-1, this is not needed.

PR #401, opened 2023-10-13
[VAULT-2589][Larky Logging]Integrate Starlark logging with Horizon

shreyoshi-mahato:
limit user messages to 1K, need file location so verbose

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/modules/utils/Reporter.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/utils/Reporter.java
@@ -25,6 +25,8 @@ public class Reporter {
   }
 
   static public void report(StarlarkThread thread, String msg, Console console) {
+    // limit the size of a string to 1K (1024 characters)
+    msg = msg.length() <= 1024 ? msg : msg.substring(0, 1024);
     if (console == null) {
       console = LogConsole.writeOnlyConsole(System.err, true);
     }
--- a/larky/src/main/java/com/verygood/security/larky/parser/DefaultLarkyInterpreter.java
+++ b/larky/src/main/java/com/verygood/security/larky/parser/DefaultLarkyInterpreter.java
@@ -43,7 +43,7 @@ public class DefaultLarkyInterpreter {
   }
 
   public ParsedStarFile evaluate(StarFile script, Writer writer) throws IOException, EvalException {
-    return evaluate(script, new StreamWriterConsole(writer));
+    return evaluate(script, new StreamWriterConsole(writer, true));
   }
 
   public ParsedStarFile evaluate(StarFile script, Console console) throws IOException, EvalException {


larky/src/main/java/com/verygood/security/larky/modules/utils/Reporter.java
```diff
@@ -25,6 +25,8 @@ public void reportError(StarlarkThread thread, String message) {
   }
 
   static public void report(StarlarkThread thread, String msg, Console console) {
+    // limit the size of a string to 1K (1024 characters)
+    msg = msg.length() <= 1024 ? msg : msg.substring(0, 1024);
```
Iapetus999, line 29: We can't do it like this after the fact. Also this needs to be configurable.

mahmoudimus: I agree with @lapetus999, this probably needs to change from a `String` to a buffer of a configurable fixed size that gets full and starts to drop messages.

larky/src/main/java/com/verygood/security/larky/parser/DefaultLarkyInterpreter.java
```diff
@@ -43,7 +43,7 @@ public DefaultLarkyInterpreter(StarlarkMode mode, Bindings... bindings) {
   }
 
   public ParsedStarFile evaluate(StarFile script, Writer writer) throws IOException, EvalException {
-    return evaluate(script, new StreamWriterConsole(writer));
+    return evaluate(script, new StreamWriterConsole(writer, true));
```
Iapetus999, line 46: As an alternative, could we add 
`evaluate(StarFile script, Writer writer, boolean includeLineNumbers)`
as a separate method? Would we be able to call that or would it require more plumbing?

Iapetus999: Or we can change the sig of this method

Iapetus999: Or we could add a flag to the bindings 🤔

shreyoshi-mahato: > As an alternative, could we add `evaluate(StarFile script, Writer writer, boolean includeLineNumbers)` as a separate method? Would we be able to call that or would it require more plumbing?

It will require changes in [LarkyCompiledScript](https://github.com/verygoodsecurity/starlarky/blob/6bc155ffd2aed4c20002638a98431d77beb5660c/larky/src/main/java/com/verygood/security/larky/jsr223/LarkyCompiledScript.java#L55)

> Or we can change the sig of this method

This would require a change in [Reporter](https://github.com/verygoodsecurity/starlarky/blob/6bc155ffd2aed4c20002638a98431d77beb5660c/larky/src/main/java/com/verygood/security/larky/modules/utils/Reporter.java#L31)

Iapetus999:
```
    String verbose = interpreter.getGlobals().getOrDefault("__VERBOSE__", "false").toString();
    return evaluate(script, new StreamWriterConsole(writer, Boolean.parseBoolean(verbose)));
```

Iapetus999: Then add 
`bindings.put("__VERBOSE__", "true")`
in `ExecutionWrapper.constructBinding()`

larky/src/main/java/com/verygood/security/larky/parser/DefaultLarkyInterpreter.java
```diff
@@ -43,7 +43,8 @@ public DefaultLarkyInterpreter(StarlarkMode mode, Bindings... bindings) {
   }
 
   public ParsedStarFile evaluate(StarFile script, Writer writer) throws IOException, EvalException {
-    return evaluate(script, new StreamWriterConsole(writer));
+    String verbose = interpreter.getGlobals().getOrDefault("__VERBOSE__", "false").toString();
```
Iapetus999, line 46: let's make a constant for `__VERBOSE__`

mahmoudimus:
@Iapetus999 have you considered using semantics instead of globals? https://github.com/verygoodsecurity/starlarky/blob/master/larky/src/main/java/com/verygood/security/larky/LarkySemantics.java#L5

Iapetus999:
Not sure what you mean. Is there an example of how this used? I'm trying to make things configurable in case we want different behavior Horizon vs Everest vs etc

Iapetus999:
"passed in" is not really a thing here. All we "pass in" are bindings. I don't see a place where semantics even get picked up, it looks like it defaults to `LarkySemantics.LARKY_SEMANTICS`

mahmoudimus:
[When a Larky Script is parsed, before it is evaluated, it can be passed Semantics that are set when you configure the host application -- in your case, it is "Everest" or "Horizon"](https://github.com/verygoodsecurity/starlarky/blob/master/larky/src/main/java/com/verygood/security/larky/parser/LarkyScript.java#L108)

If there is resistance to the idea of using semantics, that's fine, but now we have multiple options to define various behavioral toggles and in my experience, that always leads to both a maintenance issue and a source of confusion.

Iapetus999 (approved):

sharifelgamal (approved):
we can talk about Semantics more later, this will work for now

PR #422, opened 2023-12-11
[VAULT-2774][VAULT-2775] Add sign and verify to VaultModule

sharifelgamal:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
@@ -223,7 +223,7 @@ public class VaultModule implements LarkyVault {
         }
     )
     @Override
-    public Object sign(Object keyArn, Object message, Object algorithm) throws EvalException {
+    public Object sign(String keyArn, String message, String algorithm) throws EvalException {
         validateSigningAlgorithm(algorithm);
         return vault.sign(keyArn, message, algorithm);
     }
@@ -263,7 +263,7 @@ public class VaultModule implements LarkyVault {
         }
     )
     @Override
-    public Object verify(Object keyArn, Object message, Object signature, Object algorithm) throws EvalException {
+    public Object verify(String keyArn, String message, String signature, String algorithm) throws EvalException {
         validateSigningAlgorithm(algorithm);
         return vault.verify(keyArn, message, signature, algorithm);
     }
@@ -278,7 +278,7 @@ public class VaultModule implements LarkyVault {
             throw Starlark.errorf(String.format(
                 "Storage '%s' not found in supported storage types: %s",
                 storage,
-                supportedStorage.toString()
+                supportedStorage
             ));
         }
     }
@@ -293,7 +293,7 @@ public class VaultModule implements LarkyVault {
             throw Starlark.errorf(String.format(
                     "Format '%s' not found in supported format types: %s",
                     format,
-                    supportedFormat.toString()
+                    supportedFormat
             ));
         }
     }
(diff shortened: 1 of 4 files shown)


larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
```diff
@@ -188,17 +194,90 @@ public void delete(Object value, Object storage) throws EvalException {
         vault.delete(value, storage);
     }
 
+    @StarlarkMethod(
+        name = "sign",
+        doc = "signs a payload with given keyId",
+        parameters = {
+            @Param(
+                name = "keyId",
```
mjallday, line 202: should this be called `keyArn`? i think id and arn are distinct in aws platform.

https://docs.aws.amazon.com/kms/latest/developerguide/find-cmk-id-arn.html

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
```diff
@@ -188,17 +194,90 @@ public void delete(Object value, Object storage) throws EvalException {
...
+            @Param(
+                name = "message",
+                doc = "the payload to sign",
+                allowedTypes = {
+                    @ParamType(type = String.class)
+                }
+            ),
+            @Param(
+                name = "algorithm",
+                doc = "the algorith to use to sign.",
```
mjallday, line 217: do we need to include the algorithms here or make this an enum? i cannot guess what values go here so i need docs.

sharifelgamal: i'm fine adding the choices in the doc.

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
mahmoudimus: Are we exposing AWS internals too much? I understand the lack of abstraction, but would this be a breaking change if someone decided to use a different key provider?

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
mahmoudimus, line 217: spelling, should be algorithm (fyi).

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
```diff
@@ -214,25 +214,26 @@ public void delete(Object value, Object storage) throws EvalException {
             ),
             @Param(
                 name = "algorithm",
-                doc = "the algorith to use to sign.",
+                doc = "the algorithm to use to sign, valid choices are: RSASSA_PSS_SHA_256, RSASSA_PSS_SHA_384, " +
```
mjallday, line 217: 👍

mjallday:
do we have to pass [MessageType](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#API_Sign_RequestSyntax)?

sharifelgamal:
> do we have to pass [MessageType](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#API_Sign_RequestSyntax)?

We're assuming MessageType.RAW for everyone

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
sharifelgamal: we can keep it as keyId to keep it as generic as possible, but if we end up supporting HSMs outside KMS, things will get really complicated really fast

mjallday: yeah, we can add a `provider` parameter later. i see your point on `keyId`.

mahmoudimus (approved):

larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java
Iapetus999: Every place we use this in the current code base, we use `...KEY_ID`

sharifelgamal:
Moved to https://github.com/verygoodsecurity/starlarky/pull/423

PR #427, opened 2023-12-19
[VAULT-2870] removed used env variables in semgrep

shreyoshi-mahato:


Iapetus999:
starlark won't build PR's from forks anymore

PR #432, opened 2023-12-26
[VAULT-2910]remove all semgrepignore files or comments

shreyoshi-mahato:


Iapetus999:
I don't think this is the right solution

Iapetus999 (approved):

PR #441, opened 2024-01-22
WIP: give direct access to path and method for VGSHttpRequest

sharifelgamal:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/resources/vgs/http/request.star
+++ b/larky/src/main/resources/vgs/http/request.star
@@ -55,6 +55,11 @@ def VGSHttpRequest(
         # call super init, with overrides
         self.__init__(url, data=data, headers=headers, method=method)
         self.url = url
+        parsed_url = parse.urlsplit(url)
+        self.path = parsed_url.path
+        self.query_string = parsed_url.query
+        if method:
+            self.method = method
 
         return self
 


larky/src/main/resources/vgs/http/request.star
```diff
@@ -55,6 +55,11 @@ def VGSHttpRequest(
         # call super init, with overrides
         self.__init__(url, data=data, headers=headers, method=method)
         self.url = url
+        parsed_url = parse.urlsplit(url)
```
Iapetus999, line 58: what if this fails?

Iapetus999: Or can we just suggest running this code in Larky 🤔

sharifelgamal: i'm pretty sure this is guaranteed to succeed if we've already gotten this far, since we do the exact same thing inside of __init__

sharifelgamal: and yeah, this is effectively a convenience, but it's what's being asked for in the ticket

Iapetus999 (approved):

Iapetus999:
test?

sharifelgamal:
> test?

https://github.com/verygoodsecurity/starlarky/blob/master/larky/src/test/resources/vgs_tests/http/test_default_request.star#L112

PR #442, opened 2024-01-23
[PAYMTS-1728] Add account updater look up service

fangpenlin:
Add account updater look up service

## Documentation
Add account updater look up service

- [ ] Yes
- [x] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


odovzhenko (approved):

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/LarkyAccountUpdater.java
```diff
@@ -0,0 +1,31 @@
...
+   * @param expireMonth card's expiration month
+   * @param expireYear card's expiration year
+   * @param name the name on the card
+   * @param clientId client id of service account to access calm API
+   * @param clientSecret client secret of service account to access calm API
+   * @param thread Starlark thread object
+   * @return a dict contains the network token values
+   */
+  Object lookupCard(
+      String number,
```
Iapetus999, line 23: shouldn't this be `pan` or something more specific?

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/LarkyAccountUpdater.java
```diff
@@ -0,0 +1,31 @@
...
+      StarlarkInt expireMonth,
+      StarlarkInt expireYear,
```
Iapetus999, line 25: Does this have to be a 4-character string (full year)

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/LarkyAccountUpdater.java
```diff
@@ -0,0 +1,31 @@
...
+      String name,
+      String clientId,
+      String clientSecret,
```
Iapetus999, line 28: secret? 🤔

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/MockAccountUpdaterService.java
```diff
@@ -0,0 +1,39 @@
+package com.verygood.security.larky.modules.vgs.aus;
+
+import com.google.common.collect.ImmutableMap;
+import com.verygood.security.larky.modules.vgs.aus.spi.AccountUpdaterService;
+import java.util.Map;
+
+public class MockAccountUpdaterService implements AccountUpdaterService {
+
+  private static final Map<String, Card> CARDS =
```
Iapetus999, line 9: Where is `Card` defined??

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
```diff
@@ -0,0 +1,34 @@
...
+      String number,
+      Integer expireMonth,
+      Integer expireYear,
+      String name,
+      String clientId,
+      String clientSecret);
+
+  @Data
+  @Builder
+  class Card {
```
Iapetus999, line 28: consider making `record` instead of `class`

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
```diff
@@ -0,0 +1,34 @@
...
+   *
+   * @param number card's number
+   * @param expireMonth card's expiration month
+   * @param expireYear card's expiration year
+   * @param name the name on the card
+   * @param clientId client id of service account to access calm API
+   * @param clientSecret client secret of service account to access calm API
+   * @return the updated card
+   */
+  Card lookupCard(
```
Iapetus999, line 18: Not sure of the point of this method since it returns what it was passed in

Iapetus999: Or does it return a _different_ card? 🤔

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
```diff
@@ -0,0 +1,34 @@
...
+      String number,
+      Integer expireMonth,
+      Integer expireYear,
+      String name,
+      String clientId,
```
Iapetus999, line 23: wonder why these need to be passed in and why they are env vars or something

larky/src/main/java/com/verygood/security/larky/ModuleSupplier.java
```diff
@@ -20,6 +20,7 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 
+import com.verygood.security.larky.modules.AccountUpdaterModule;
```
mahmoudimus, line 23: This does not belong in the general modules package. This belongs in the `vgs` package.

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
```diff
@@ -0,0 +1,123 @@
...
+            String.format(
+                "AccountUpdaterModule expecting only 1 network token provider of type AccountUpdaterService, found %d",
+                AccountUpdaterProviders.size()));
+      }
+      AccountUpdaterService = AccountUpdaterProviders.get(0);
+    }
+  }
+
+  @StarlarkMethod(
+      name = "lookup_card",
```
mahmoudimus, line 50: should this change to lookup PAN instead of lookup card to match the documentation?

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
```diff
@@ -0,0 +1,123 @@
...
+      doc = "Lookup account updates for a given PAN.",
+      useStarlarkThread = true,
+      parameters = {
+        @Param(
```
mahmoudimus, line 54: Are all of these parameters required? If we pass `None`, this code will throw an error. Is this the desired behavior?

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
```diff
@@ -0,0 +1,123 @@
...
+      throws EvalException {
+    if (number.trim().isEmpty()) {
+      throw Starlark.errorf("card number argument cannot be blank");
+    }
+    final AccountUpdaterService.Card card;
+    try {
+      card =
+          AccountUpdaterService.lookupCard(
+              number,
+              expireMonth.toInt("invalid int range"),
```
mahmoudimus, line 106: Let's make this more user friendly - maybe put `expireMonth` in the error?

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
```diff
@@ -0,0 +1,123 @@
...
+              expireYear.toInt("invalid int range"),
```
mahmoudimus, line 107: Same

mahmoudimus (changes requested):

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
```diff
@@ -0,0 +1,123 @@
...
+            doc =
+                "Card expiration year. Used to pass to the network for retrieving the corresponding card information",
+            allowedTypes = {@ParamType(type = StarlarkInt.class)}),
+        @Param(
+            name = "name",
+            named = true,
+            doc =
+                "Card owner name. Used to pass to the network for retrieving the corresponding card information",
+            allowedTypes = {@ParamType(type = String.class)}),
+        @Param(
```
mahmoudimus, line 77: For me, this immediately stood out as 'hmm.' as the credentials are passed in via method call. Is this what we are intending?

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
fangpenlin: I tried to put record syntax here, but my IDE is showing red font. It seems like record is supported only in Java 14, are we using Java 14 in this larky project?

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
fangpenlin: yeah, those are meant to be required, so passing None will result in error, that's desired behavior

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
fangpenlin: Yes, that's what we are intending.

This is for our needs in payments team to make it possible for our customer to get the latest card number from account updater service (provided by visa or mastercard). And to achieve that, we need to make a call to public API of calm. I envisioned this could be used like this

```python
if use_account_updater(input.headers):
    card = aus.lookup_card(
        pan=body.cardNumber,
        exp_month=body.expMon,
        exp_year=body.expYear,
        name=body.name,
        client_id="<CALM SERVICE ACCOUNT CLIENT ID>",
        client_secret="<CALM SERVICE ACCOUNT CLIENT SECRET>",
    )
    if card is not None and card.number != input.cardNumber or ...:
        input.cardNumber = card.number
        # update card
# proceed to PSP with or without new card info
```

As you can see the challenge here is to authenticate with calm API server. The easiest way right now is to hardwire the service account credentials in the larky script. I know this may not be the best way in the world to authenticate to our own API service in VGS, but considering we need this feature in short time, it's the best way. If there's a concern for hardwiring the credentials in the larky script, we can actually ask the merchant to put their service account credentials in the vault and use `reveal` function to get the actual credential secret values. Like this


```python
card = aus.lookup_card(
     # ...
    client_id=vault.reveal("<CALM SERVICE ACCOUNT CLIENT ID>"),
    client_secret=vault.reveal("<CALM SERVICE ACCOUNT CLIENT SECRET>"),
)
```

And if we don't want the possibility of merchant putting their service account credentials in larky script and force them to use vault, we can even reveal the secret in module and reject the request if the provided values are not alias.

In a perfect world, if we have enough time to build this feature and some major changes to vault is allowed, I was thinking about providing something like OIDC auth token like the ones from Kubernetes pod for authenticating with Calm server. Since the larky script environment is kind of already a trusted environment, I don't see why we need to authenticate again. So requests coming from larky can be trusted as long as there's a policy says that we trust calm API requests coming from our own vault proxy. Then the API call can be simplified as just 

```python
card = aus.lookup_card(
    pan=body.cardNumber,
    exp_month=body.expMon,
    exp_year=body.expYear,
    name=body.name,
)
```

The account updater module will create a OIDC auth token for making the request and calm can authenticate and accept that automatically based on a pre-defined policy.

Regardless, that's too much effort for now, as we want this feature to be done ASAP, provides if keeping calm's service account credentials in vault (or even hardwiring in larky script) is acceptable, I think this is the best approach for now.

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/LarkyAccountUpdater.java
fangpenlin: please see my comment at https://github.com/verygoodsecurity/starlarky/pull/442#discussion_r1463854857 explaining the rationale

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
williamohara: To me the ability to use account updater should be done through a module. I don't feel that we should be using credentials like this for service to service calls. Here is my thinking.

Create a Vault Preference called "Account Updater" that is default false.
If that preference is false - horizon, simply does not allow the call to the service be completed.
If that preference is true - it goes through

The preference can only be set by someone/something with permissions to do so. ( this is something that the new Vault Management spec intends to support)

fangpenlin: so it sounds like we only want to enable this feature for customers who really need it, is my understanding correct?

is this vault preference feature something we can use right now, or is this a new feature and not yet implemented? I did a quick search in the code base and cannot find something looks like vault preference. If it's already there, can you please point that to me where's it and if there's an example of it?

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
fangpenlin: A card could be expired, exp date changed, or stolen, so the number changed. It's still the same card, but a different number or exp date. So this service and this method is for looking up the given card number and see if there's updates for the card. The customer would like to use the new card number is there's one in case of the original one is expired.

larky/src/main/java/com/verygood/security/larky/modules/AccountUpdaterModule.java
williamohara: Yes, it's there... @sharifelgamal or @Iapetus999 can point you in the right direction. We made it so that preferences for the vault are available in horizon. Adding a new preference - 🤷🏼  not sure about. but I am going to hope it can be quick to add one🤞🏼 .

sharifelgamal: note that we never actually added the ability to read preferences from horizon, it would involve the proxy (or operation pipeline as it were) passing in the preference as a header into larky

larky/src/main/java/com/verygood/security/larky/modules/vgs/aus/spi/AccountUpdaterService.java
Iapetus999: You should be using JDK 17 here

PR #454, opened 2024-02-12
Update Python, JVM in build

Iapetus999:
- Update Build to use JVM 17 target code
- Update Python to 3.10.x

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No


sharifelgamal (approved):
seems reasonable

PR #464, opened 2024-03-06
Return step stats

Iapetus999:
- Change CPU Ticks to `long`
- Change snapshot version to `1.0.0-SNAPSHOT`
- Add execution steps and cpu ticks to module globals on exit.

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/pom.xml
+++ b/larky/pom.xml
@@ -7,7 +7,7 @@
 
     <groupId>com.verygood.security</groupId>
     <artifactId>larky</artifactId>
-    <version>1.0-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


sharifelgamal (approved):

mahmoudimus:
Can we discuss modification of LibstarLark ?

Iapetus999:
All I did was change an `int` to a `long` for CPU ticks and then added a getter for it.
It only works on certain platforms so I wasn't able to test it. I don't think we'll wind up using it but it seems like an obvious change.

PR #479, opened 2024-04-04
VAULT-3067 Add time limit

Iapetus999:
- Adds a time limit (expiration) to evaluator.
- Allows setting params to specify step limit and expiration

### DIFF
--- a/libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
+++ b/libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
@@ -14,13 +14,16 @@
 
 package net.starlark.java.eval;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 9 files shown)


sharifelgamal (approved):

libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
```diff
@@ -86,6 +91,54 @@ public void setMaxExecutionSteps(long steps) {
...
+   *
+   * @return step limit
+   */
+  public long getStepLimit() {
+    return stepLimit;
+  }
+
+  /**
+   *
+   * @return expiration date in ms since 1970
```
mjallday, line 104: this is surprising to me. i thought ms would be relative not, absolute since 1970

mjallday: having it be absolute would mean a job could not be re-run without recomputing the value.

Iapetus999:
Replacing with https://github.com/verygoodsecurity/atlas-horizon/pull/419

PR #489, opened 2024-04-17
Add step limit and expiration redux

Iapetus999:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
+++ b/libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
@@ -14,13 +14,16 @@
 
 package net.starlark.java.eval;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


libstarlark/src/main/java/net/starlark/java/eval/StarlarkThread.java
```diff
@@ -86,6 +91,54 @@ public void setMaxExecutionSteps(long steps) {
...
+   */
+  public Clock getClock() {
+    return clock;
+  }
+
+  /**
+   * Sets a specific clock object for testing purposes
+   * @param clock Clock for testing
+   */
+  @VisibleForTesting
```
sharifelgamal, line 133: is this used anywhere?

Iapetus999: No but it would let a test case set a mock clock impl (since we don't have a formal injection framework in Starlark).

sharifelgamal (approved):

PR #545, opened 2024-10-09
WIP: [PYMTPLAT-2987] Add transaction type for network token cryptogram request

fangpenlin:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [ ] No

- [ ] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
@@ -92,6 +92,11 @@ public class NetworkTokenModule implements LarkyNetworkToken {
             doc = "VGS merchant id to get a network token for",
             defaultValue = "''",
             allowedTypes = {@ParamType(type = String.class)}),
+        @Param(
+            name = "transaction_type",
+            named = true,
+            doc = "Type of payment transaction for requesting cryptogram",
+            allowedTypes = {@ParamType(type = String.class)}),
       })
   @Override
   public Dict<String, Object> getNetworkToken(
@@ -101,16 +106,18 @@ public class NetworkTokenModule implements LarkyNetworkToken {
       String currencyCode,
       String cryptogramType,
       String vgsMerchantId,
+      String transactionType,
       StarlarkThread thread)
       throws EvalException {
     if (pan.trim().isEmpty()) {
       throw Starlark.errorf("pan argument cannot be blank");
     }
     final Optional<NetworkTokenService.NetworkToken> networkTokenOptional;
+    // TODO: handle txn type here
     try {
       networkTokenOptional =
-          networkTokenService.getNetworkToken(pan, cvv, amount, currencyCode, cryptogramType,
-              vgsMerchantId);
+          networkTokenService.getNetworkToken(
+              pan, cvv, amount, currencyCode, cryptogramType, vgsMerchantId, transactionType);
     } catch (UnsupportedOperationException exception) {
       throw Starlark.errorf("nts.get_network_token operation must be overridden");
     }
--- a/larky/src/main/java/com/verygood/security/larky/modules/vgs/nts/spi/NetworkTokenService.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/vgs/nts/spi/NetworkTokenService.java
@@ -13,10 +13,17 @@ public interface NetworkTokenService {
    * @param amount amount of payment for retrieving cryptogram
    * @param currencyCode currency code of payment for retrieving cryptogram
    * @param cryptogramType type of cryptogram
+   * @param transactionType type of transaction for requesting cryptogram
    * @return the network token value
    */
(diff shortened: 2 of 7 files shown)


larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
```diff
@@ -101,16 +110,32 @@ public Dict<String, Object> getNetworkToken(
       String currencyCode,
       String cryptogramType,
       String vgsMerchantId,
+      @Nullable Object transactionType,
```
Iapetus999, line 113: Why not just make it `String`?

larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
```diff
@@ -101,16 +110,32 @@ public Dict<String, Object> getNetworkToken(
...
+                    .cvv(cvv)
+                    .amount(amount)
+                    .currencyCode(currencyCode)
+                    .cryptogramType(cryptogramType)
+                    .merchantId(vgsMerchantId)
+                    .transactionType(
+                        transactionType instanceof String ? (String) transactionType : null)
+                    .build());
+      } else {
+        networkTokenOptional =
```
Iapetus999, line 135: When would this be true?

larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
fangpenlin: Because it's possible the passed in parameter is a `None` in larky, i.e:

`net.starlark.java.eval.NoneType`

And if it's not the case, it would be a `String`.

I checked other larky methods with potential None value, and it appears they all use `Object` type. Take vault module's `storage` parameter for example, it's also an object:

https://github.com/verygoodsecurity/starlarky/blob/81e44ad1b1b528b5557e4d1456d24f5fcd0b92a5/larky/src/main/java/com/verygood/security/larky/modules/VaultModule.java#L142

larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
fangpenlin: When the provided `networkTokenService` object is just a `NetworkTokenService` rather than `NetworkTokenServiceV2`. I have concerns regarding if we make a breaking change here in starlarky and make a new release, atlas-commons and atlas-horizon will get stuck with legacy version of starlarky. Unless, the new `NetworkTokenService` is implemented in atlas-commons immediately after this PR merged and release.

So breaking up `NetworkTokenService` and `NetworkTokenServiceV2`, making `NetworkTokenService` still workable is just to ensure even if we merge this and release new larky right now, it shouldn't break `atlast-commons` when upgrading larky without implementing with the new method signature.

If that's not a concern, I can just drop the two interfaces approach and change the `NetworkTokenService` interface directly. Please let me know what you think.

larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
Iapetus999: Bah, wish they'd come up with something else like `Optional` this is annoying

larky/src/main/java/com/verygood/security/larky/modules/NetworkTokenModule.java
Iapetus999: Not sure I follow. You can update Larky, release it, update the parent pom, then apply the new parent pom to atlas-commons, release that (which updates PP automatically) then apply that new PP to horizon/everest

fangpenlin: The only concern I have is that if we merge this PR, and somehow cannot update `atlast-horizon` in time, other releases will be blocked until the another PR is done. But since as if this is not a concern and we are going to fix the stuff and merge it quickly, I am going to simplify the code by using just one interface with new signature then.

fangpenlin: fixed, now I only use one interface. I will have `atlast-commons` PR coming up soon

fangpenlin:
@Iapetus999  can you please review again?

larky/src/main/java/com/verygood/security/larky/modules/vgs/nts/spi/NetworkTokenService.java
```diff
@@ -1,22 +1,38 @@
 package com.verygood.security.larky.modules.vgs.nts.spi;
 
 import java.util.Optional;
+import javax.annotation.Nullable;
```
Iapetus999, line 4: This is an aside but we need to move everything to `jakarta.`

Iapetus999 (approved):

PR #570, opened 2025-01-15
[CSL3-2519] Use correct IV length for AES CBC

viacheslav-rud:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/test/resources/vendor_tests/jose/test_jose.star
+++ b/larky/src/test/resources/vendor_tests/jose/test_jose.star
@@ -307,6 +307,71 @@ def test_encrypt_and_decrypt_with_certificate():
     
     asserts.assert_that(decrypted).is_equal_to(payload)
 
+def test_encrypt_and_decrypt_with_certificate_AES_CBC():
+    certificate = """-----BEGIN CERTIFICATE-----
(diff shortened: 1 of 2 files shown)


larky/src/test/resources/vendor_tests/jose/test_jose.star
```diff
@@ -307,6 +307,71 @@ def test_encrypt_and_decrypt_with_certificate():
     
     asserts.assert_that(decrypted).is_equal_to(payload)
 
+def test_encrypt_and_decrypt_with_certificate_AES_CBC():
```
Iapetus999, line 310: Hate to say this but SAST isn't going to be happy with this. Not sure what the alternative is

Iapetus999 (approved):

larky/src/test/resources/vendor_tests/jose/test_jose.star
viacheslav-rud: It’s been already there in a different test. The test I added is basically duplicate of another existing test but with changed algorithm

mjallday (changes requested):
This code doesn’t mirror the original change from the upstream repo. Please help me understand why

Agitolyev (approved):

PR #578, opened 2025-03-12
[PROXY-2510] Larky OpenPGP SHA384 Fixes

kochniev:
1. Fix the methods below would return an error:
 - `OpenPGP.enarmor(encrypted_body.to_bytes(), marker='MESSAGE', headers = None)`
 - `sign.sign(data, hash="SHA384")`
2. Change the hardcoded date used to generate the signature from a day in 2022 to March 12, 2029. This is a temporary fix to unblock signature generation

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No


kochniev:
I spent several hours wrestling with the error below. I reverted all commits in this branch to the point where it was identical to master, and I was still getting the same error. 
I couldn't figure out the root cause. Maybe it was because I pushed to my fork and upstream. I'm not sure. But [recreating](https://github.com/verygoodsecurity/starlarky/pull/583) the PR and pushing only to upstream seems to have helped. 

image

PR #608, opened 2025-04-30
[CSL3-2693] Add parameters to choose key for operations

viacheslav-rud:
(insert text here)

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/main/java/com/verygood/security/larky/modules/vgs/PGPModule.java
+++ b/larky/src/main/java/com/verygood/security/larky/modules/vgs/PGPModule.java
@@ -7,13 +7,10 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
-import java.util.ArrayList;
 import java.util.Date;
+import java.util.HexFormat;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Spliterator;
-import java.util.Spliterators;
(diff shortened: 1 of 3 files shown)


larky/src/main/java/com/verygood/security/larky/modules/vgs/PGPModule.java
```diff
@@ -719,6 +725,18 @@ private PGPPublicKey readPublicKey(String armoredKey)
...
+        while (signatures.hasNext()) {
+            PGPSignature signature = signatures.next();
+            PGPSignatureSubpacketVector subpacketVector = signature.getHashedSubPackets();
+            if (subpacketVector.getKeyFlags() == (KeyFlags.ENCRYPT_COMMS | KeyFlags.ENCRYPT_STORAGE)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
```
Iapetus999, line 739: test?

viacheslav-rud: Existing test already failed so it is tested, I'll add a comment to the test that encryption subkey needs to have E flag

viacheslav-rud: We changed this logic, tests were also added

Agitolyev (approved):

PR #669, opened 2025-11-23
Fixed bug in randrange method. Added tests for random library.

GKotsifos:
1. Fixed issue on randrange to properly accept 1 and 2 arguments
2. Added comprehensive test suite for the random stdlib

## Documentation
(insert text here)

- [ ] Yes
- [x] No

- [x] Use previous release
- [ ] Use a feature flag
- [ ] No

### DIFF
--- a/larky/src/test/resources/stdlib_tests/test_random.star
+++ b/larky/src/test/resources/stdlib_tests/test_random.star
@@ -1,25 +1,27 @@
 load("@stdlib//random", random="random")
 load("@vendor//asserts", "asserts")
 load("@stdlib//unittest", "unittest")
+load("@stdlib//larky", "larky")
 
 # randrange tests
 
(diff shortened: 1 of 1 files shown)


aslepakurov:
I think, yeah, the set up is to run CircleCI tests not in fork
@mjallday can you please help

larky/src/test/resources/stdlib_tests/test_random.star
```diff
@@ -0,0 +1,210 @@
...
+
+
+def test_getrandbits_negative_fails():
+    random.getrandbits(-1)
+
+
+# randbytes tests
+
+def test_randbytes_basic():
+    b = random.randbytes(16)
```
mahmoudimus: can you assert that this is of type bytes?

